PR #65, opened 2026-06-03
[PR-12-code] Migrate to neoipcr post-schema-arc API; revive Validation Report; align PowerShell wrappers

Brar:
First half of the post-schema-arc Surveillance-Toolkit catch-up. Code-only — `.po`/`.pot`/`_sR.yaml` files split into [PR-12-translations] (will open after this merges; ~40 files / ~200k lines of machine-derived regeneration triggered by the API migration here).

## Scope

Surveillance-Toolkit catches up with neoipcr's PR-09 + PR-10 + PR-11 stack.

### neoipcr API migration
- All `dhis2_dataset_options()` call sites migrate to the post-Phase-A/B/C shape (`pseudo` / `full` option values; three-mode link-privacy gates; `patient_columns = "id"` instead of the legacy `include_patient_id`; `patient_columns = "gestational_age"` instead of `gest_age`).
- Patient-Data-Report: migrate from `ds\$eventDetails` to `ds\$events` (Phase B fold-in).
- Validation Report: revived for the current neoipcr API shape; `hospital_key` join through `metadata\$departments` per the schema-arc reader-owned narrowing.
- Universal guardrail added: **never join neoipcr tibbles on DHIS2 UIDs** (use synthesized integer keys).

### Validation Report build-out
- Antibiotic utilisation table + sparse-data footnote.
- Exclude test data by default; add `-IncludeTestData` switch.
- Rename `New-ValidationReports.ps1` → `New-ValidationReport.ps1` (singular).

### PowerShell wrapper alignment
- Align variable naming across all `scripts/New-*.ps1` wrappers.
- Add the wrapper-script alignment guardrail (repo-specific).
- Partner Report: add `-ValidationExceptionFile` and `-IncludeTestData`.
- Sync Partner-Report param defaults across languages.
- `Get-NeoipcDepartments`: add `-ExcludeTestUnits` switch.
- `helpers.R`: restore comment on `format_countries` name expectation.
- `load_neoipcr`: honor `NEOIPCR_DEV_PATH` env-var override.

### Universal guardrails sync
- Verify-upstream-now, verify-factual-claims, re-read-iteratively, dismiss-inconsistencies, filler-comments, predict-future, NotImplementedException, Co-Authored-By, long-lived-guidance.
- Doc-comments in-band (not as a sweep).
- No deprecated APIs + read the source.
- **Namespace-qualify recommended-package calls** (universal-tier, propagated here from neoipcr's PR-09 round 2 review fix that landed after the dev branch began).

## Audit re-applied
- `reports/Validation-Report/_content.qmd:29` — `paste(..., sep = ", ")` → `paste(..., collapse = ", ")`. Cherry-picks the PR-07a-code round 12 production-tip fix that the dev branch never absorbed (rendered the department-code list as space-separated values instead of `A, B, C`).

## Pre-flight

155 files / 8,159 lines (under both Copilot review caps of 300 files / 20,000 lines). Translation artefacts (40 files / ~208k lines, all machine-derived from sources reviewed here) split into [PR-12-translations].

## Verification

- All R / YAML files parse cleanly (no Quarto reports rendered — DHIS2 API guardrail).
- The audit re-applied fix is single-line and visually inspected; no other PR-07a-code/PR-07b-code review fixes deviate from the dev tip (verified via `tmp/upstreaming-review-log.md`).

### Blocks on (must merge first)
- [PR-07b-translations] NeoIPC/Surveillance-Toolkit#64 ✅ MERGED
- [PR-10] NeoIPC/neoipcr#6 ✅ MERGED (schema arc — provides `pseudo`/`full` rename, three-mode gates, `gestational_age` rename, `assert_data_protection`)
- [PR-11] NeoIPC/neoipcr#7 ✅ MERGED (`write_json` — used downstream of the reference-data sidecar, not in this PR)

### Blocked by this PR
- [PR-12-translations] — opens after this merges (GitHub doesn't support stacked fork→upstream bases).
- [PR-13] — ST metadata hygiene, machine-readable params, NeoIPC-Tools cmdlets, deprecation sweep.


.github/workflows/build.yml
```diff
@@ -14,26 +14,20 @@ jobs:
...
             - name: Run po4a
               shell: bash
               run: |
-                # Use the pinned po4a submodule (tools/po4a) — the Ubuntu apt
-                # po4a is too old to recognise the --width / --wrap-po options
-                # that our po4a configs use. Invocation pattern matches the
-                # Surveillance-Toolkit CLAUDE.md "po4a setup" section.
-                PERLLIB=tools/po4a/lib tools/po4a/po4a --verbose po/documentation.po4a.cfg
-                PERLLIB=tools/po4a/lib tools/po4a/po4a --verbose po/infectious_agents.po4a.cfg
+                po4a --verbose po4a.cfg
```
Brar: 🔴 **CI regression — this reverts the po4a step to a broken state.** Base `main` runs po4a via the pinned `tools/po4a` submodule with the real config paths under `po/`; this hunk drops `submodules: true`, switches to apt `po4a`, and invokes:
```
po4a --verbose po4a.cfg
po4a --verbose metadata/common/infectious-agents/po4a.cfg
```
**Neither path exists** — the configs are `po/documentation.po4a.cfg`, `po/infectious_agents.po4a.cfg`, `po/reports.po4a.cfg`. So the CI po4a step will fail outright, and it also contradicts the repo's own CLAUDE.md po4a guidance ("apt po4a is too old… use the pinned submodule"). This looks like a stale-base/revert hunk from the dev branch clobbering the merged-main version rather than intended content. Please restore the base's version (pinned submodule + `po/*.po4a.cfg` paths + `submodules: true` + the gettext/perl deps).

---

reports/Validation-Report/de/_problem_detail_0002.Rmd
```diff
@@ -2,4 +2,4 @@
 
 Wenn Sie das Surveillance-Ende-Formular abschließen sollten Sie normalerweise auch die Patientenregistrierung abschließen um sie im System als vollständig und fertig für die Auswertung zu markieren und sie vor versehentlichen Änderungen zu schützen.
 Wenn Sie das nicht tun, werden die Daten des betreffenden Patienten nicht in Ihren Auswertungen enthalten sein.
-Am einfachsten geht das, indem Sie im Dialogfeld, das nach dem Klicken auf „Complete“ im Surveillance-Ende-Formular angezeigt wird, auf „Complete event and enrollment“ klicken. Wenn Sie eine Patientenregistrierung im Nachhinein abschließen möchten und nicht wissen, wie das funktioniert, folgen Sie bitte der Beschreibung in @sec-solution-4.
+Am einfachsten geht das, indem Sie im Dialogfeld, das nach dem Klicken auf „Complete“ im Surveillance-Ende-Formular angezeigt wird, auf „Complete event and enrollment“ klicken. Wenn Sie eine Patientenregistrierung im Nachhinein abschließen möchten und nicht wissen, wie das funktionert, folgen Sie bitte der Beschreibung in @sec-solution-4.
```
Brar: 🔴 **This is a po4a-generated file — it shouldn't be hand-edited here.** `reports/Validation-Report/de/_problem_detail_0002.Rmd` is a po4a target (`po/reports.po4a.cfg:62` → `en/_problem_detail_0002.Rmd $lang:.../de/_problem_detail_0002.Rmd`), and CLAUDE.md lists `Validation-Report/<lang>/` among the generated dirs that must never be edited directly. The edit will be silently reverted on the next `Invoke-Localization.ps1 -Update`, and the correct path is to change the `en/` master + `.po`.

This also **introduces a typo**: `funktioniert` → `funktion**e**rt`.

Same applies to the sibling edits/deletions in this PR: `de/_problem_detail_0011/0012/0014/0020.Rmd`, `de/_solution_0001.Rmd` (modified), and `de/_solution_0003/0004/0007/0011/0015.Rmd` (deleted) — those deletions will just be regenerated. Per the PR's own "translations split into PR-12-translations" framing, none of these localized `de/` content files should be in this code PR. (NB: this is distinct from the top-level `*-Report.<lang>.qmd` profile files, which are *not* po4a-generated and whose param edits are fine.)

---

reports/Validation-Report/en/_problem_detail_0014.Rmd
```diff
@@ -7,9 +7,9 @@ The disadvantage of this decision is that none of our partners can use their fam
 
 Please always enter valid ICHI codes in the ‘Main procedure code (ICHI)’ and ‘Side procedure code (ICHI)’ fields.
 
-You can find these codes by entering the English name of the procedure (or part of it) in the ‘Search’ field at the top left of the ICHE Reference Classification at https://icd.who.int/dev11/l-ichi/en.
+You can find the these codes by entering the English name of the procedure (or part of it) in the ‘Search’ field at the top left of the ICHE Reference Classification at https://icd.who.int/dev11/l-ichi/en.
```
Brar: Typo regression introduced by this PR (`these` → `the these`):
```suggestion
You can find these codes by entering the English name of the procedure (or part of it) in the ‘Search’ field at the top left of the ICHE Reference Classification at https://icd.who.int/dev11/l-ichi/en.
```
While here: `ICHE` (this line and the "enter the ICHE code" line below) looks like it should be **ICHI** — the URL is `l-ichi` and the classification is ICHI. That's pre-existing, not introduced here, but worth fixing in the same pass since you're editing this source.

---

reports/Validation-Report/en/_problem_detail_0014.Rmd
```diff
@@ -7,9 +7,9 @@ The disadvantage of this decision is that none of our partners can use their fam
...
 
 If you have any problems finding the correct code, please contact us and we will be happy to help.
 
 To correct the reported error, please open the form (see @sec-solution-15) and enter the ICHE code for the procedure instead of the existing code.
-If the form has already been completed, it will need to be reopened for editing (see @sec-solution-9). If the patient registration has also already been completed, you will also need to reopen it for editing beforehand (see @sec-solution-5).
+If the form has already been completed, it will need to bebe reopened for editing (see @sec-solution-9). If the patient registration has also already been completed, you will also need reopen it for editing beforehand (see @sec-solution-5).
```
Brar: Two typo regressions introduced by this PR on this line (`be` → `bebe`, and `need to reopen` → `need reopen`):
```suggestion
If the form has already been completed, it will need to be reopened for editing (see @sec-solution-9). If the patient registration has also already been completed, you will also need to reopen it for editing beforehand (see @sec-solution-5).
```

---

Brar (commented):
## Review — [PR-12-code] neoipcr post-schema-arc API migration

The core migration is **clean and consistent** — I verified each call site and the data-protection-sensitive bits:

- ✅ **API value migration** — no leftover `pseudonymised`, `include_patient_id`, or `include_* = "yes"`; `pseudo`/`full`, `patient_columns = "id"`, and `gestational_age` are used correctly everywhere.
- ✅ **`eventDetails` → `events`** (only explanatory comments reference the old name).
- ✅ **New conditional tables** (organism-resistance, antibiotic-utilisation) are gated via `.show(params$…)`, and the cross-references obey the conditional-reference guardrail — the external `@tbl-organism-resistance-rate` ref in `_setup.qmd:1451` is wrapped in `if (params$includeOrganismResistanceRateTable)`, and the intro's self-reference is nested inside the same gate.
- ✅ **Namespace-qualify**, the `paste(sep=)` → `paste(collapse=)` audit fix, **CLAUDE.md ↔ copilot-instructions.md sync** (byte-identical guardrail bullets), and the `New-ValidationReports` → `New-ValidationReport` rename (zero stale references) all check out.
- ✅ The localized **`*-Report.<lang>.qmd` param edits are fine** — those are Quarto *profile* files (`_quarto.yml: profile.group`), **not** po4a targets (no `.qmd` in `po/reports.po4a.cfg`), so syncing `includeOrganismResistanceRateTable:` across them is correct, not a guardrail breach.

Four things to fix, two of them blocking:

### 🔴 CI regression — `.github/workflows/build.yml`
The po4a step is reverted to apt `po4a` with config paths that **don't exist** (`po4a.cfg`, `metadata/common/infectious-agents/po4a.cfg` — the real ones are under `po/`), and drops `submodules: true`. This breaks CI and contradicts the repo's po4a guidance. Looks like an accidental stale-base hunk. Inline.

### 🔴 po4a-generated files edited in `Validation-Report/de/`
Distinct from the profile files above: `de/_problem_detail_*.Rmd` / `de/_solution_*.Rmd` **are** po4a targets (`po/reports.po4a.cfg:58-95`). This PR modifies five `de/_problem_detail_*` + `de/_solution_0001`, deletes five `de/_solution_*`, and introduces a typo (`funktioniert`→`funktionert`). These will be clobbered by the next localization run and, per the PR's own "translations → PR-12-translations" split, shouldn't be here. Edit `en/` + `.po` instead. Inline.

### 🟡 Typo regressions in the English master `en/_problem_detail_0014.Rmd`
This PR introduces `the these`, `bebe`, and `need reopen it` into source text. Inline suggestions provided (plus a pre-existing `ICHE`→`ICHI` worth fixing in passing).

### 🟡 Wrapper-alignment gap (pre-existing, flagging per the new guardrail)
`New-PartnerReports.ps1` stays plural and uses `$authForEnv` while the other `New-*` wrappers use the aligned `$auth` shape. Not introduced here, but the wrapper-alignment guardrail this PR adds now covers it — worth a follow-up.

Net: the migration itself is solid; the two 🔴 items read as dev-branch baggage that clobbered merged-main improvements (the CI po4a step and the generated `de/` files) — worth a sweep for any other stale-base hunks in the squash. (Submitting as a comment — GitHub blocks an approving review on one's own PR.)

---

Brar:
## Correction + a render-breaking issue it surfaced

**Correction to my review:** I called the top-level `*-Report.<lang>.qmd` files "Quarto profile files" — that's wrong, and I'm striking it. Quarto profile files are `_quarto-<lang>.yml`; the `Reference-Report.de.qmd` etc. are **hand-maintained per-language report entrypoint documents** (each sets `lang:` + `LC_ALL`, carries the full `params:` block, and `{{< include _content.qmd >}}`). The *conclusion* still holds — they're not po4a-generated (no `.qmd` in `po/reports.po4a.cfg`), so syncing their `params:` across languages is legitimate and not a guardrail breach — but the label was wrong. Thanks for the catch.

**Re-checking them, though, turned up a render-breaking bug.** The new `includeAntibioticUtilisationTable` param was added to the English master `Reference-Report.qmd` but is **missing from all five localized entrypoints** — `Reference-Report.{de,el,es,et,it}.qmd` each received `includeOrganismResistanceRateTable` but **not** `includeAntibioticUtilisationTable`, and no `_quarto*.yml` supplies it as a fallback. `_content.qmd:343` calls `.show(params$includeAntibioticUtilisationTable)`, and `.show()` (in `_setup.qmd:294`) is:

```r
.show <- function(flag) cat('::: {.content-', if (flag) 'visible' else 'hidden', ...)
```

With the param undeclared, `params$includeAntibioticUtilisationTable` is `NULL`, so `if (NULL)` throws **`argument is of length zero`** — rendering any non-English Reference Report aborts at that section. (Verified the `.show(NULL)` error on R 4.3.) Partner-Report is fine (both `.qmd` and `.de.qmd` carry both new params); the gap is specific to the five Reference-Report localized variants.

This is exactly the class of bug the "no reports rendered (DHIS2 guardrail)" verification can't surface. **Fix:** add `includeAntibioticUtilisationTable: false` to those five files to match the master (`Reference-Report.qmd:39`).

---

Brar:
Followed the wrapper path to be sure the bug isn't masked there (and agreed on the premise — Quarto params can only come from the entry `.qmd`, not `_quarto*.yml`, so the per-language entry files have to replicate them).

`New-ReferenceReport.ps1` confirms it's **not** masked:
- `$qmdParams = @{}` starts empty (L263).
- The `include*Table` flags are added to `$qmdParams` **only** via `-EnableElements` / `-DisableElements`, and both default to `@()` (L106, L131).
- The render loop emits `-P` only for keys present in `$qmdParams` ("All parameters via -P" loop), so a **default** render passes no `-P includeAntibioticUtilisationTable` and falls back to the entry-file frontmatter.

So the default localized render path (`Reference-Report.{de,el,es,et,it}.qmd`, which lack the param) hits `.show(NULL)` → `argument is of length zero` and aborts. The only way it'd survive is an explicit `-EnableElements AntibioticUtilisationRates` / `-DisableElements AntibioticUtilisationRates`, which would emit the `-P`. English is unaffected (its frontmatter declares the param `false`), and Partner-Report is unaffected (both its entry files already declare both new params).

Fix stands: replicate `includeAntibioticUtilisationTable: false` into the five localized `Reference-Report` entry files to match `Reference-Report.qmd:39`.

---

.github/workflows/build.yml
Brar: Fixed in df7a9dd — restored .github/workflows/build.yml to upstream/main. The dev branch had reverted to the broken apt-po4a state; the upstream version (pinned tools/po4a submodule + po/*.po4a.cfg paths + submodules: true + the gettext/perl deps) is back.

reports/Validation-Report/de/_problem_detail_0002.Rmd
Brar: Fixed in df7a9dd — restored all eleven po4a-targeted de/ files to upstream/main (the six modified _problem_detail / _solution_0001 files reverted; the five deleted _solution files brought back). Including the funktioniert -> funktionert typo, which was in the about-to-be-clobbered de/ version anyway. Translatable content edits will go through the .po pipeline in PR-12-translations.

reports/Validation-Report/en/_problem_detail_0014.Rmd
Brar: Fixed in df7a9dd — applied the suggestion (`the these` -> `these`) plus the pre-existing `ICHE` -> `ICHI` on both this line and `the ICHE code` further down. The URL is `l-ichi` and the classification is ICHI; the previous wording came from the pre-NeoIPC-protocol era when ICHE was the working name.

reports/Validation-Report/en/_problem_detail_0014.Rmd
Brar: Fixed in df7a9dd — applied the suggestion (`bebe` -> `be`, `need reopen it` -> `need to reopen it`).

Brar:
## Round 1 responses

Five items addressed in `df7a9dd`:

**Render-breaking missing localized `includeAntibioticUtilisationTable`.** Confirmed the failure mode (English master `Reference-Report.qmd:39` has it, the five localized entry files don't, `.show()` hits `if (NULL)` → `argument is of length zero`). And confirmed your `New-ReferenceReport.ps1` trace — the default path passes no `-P includeAntibioticUtilisationTable` and falls back to the entry-file frontmatter. Fix landed: `includeAntibioticUtilisationTable: false` added to all five `Reference-Report.{de,el,es,et,it}.qmd` files at the same position the master has it (between `includeRiskDensityRateTable` and `includeSurgicalProcedureRateTable`). Inline-thread replies on the other four items reference this commit too.

**Wrapper-alignment gap — `New-PartnerReports.ps1`.** Filed as [`tasks/ps-wrapper-new-partnerreports-align.md`](tasks/ps-wrapper-new-partnerreports-align.md): rename to singular, switch from `\$authForEnv` to the aligned `\$auth` shape, audit the rest of the script against the alignment guardrail, sweep all references. Pre-existing per the review, surfaced by the new repo-specific guardrail this PR adds.

**Stale-base sweep — looked for other regressions of the same class.** Compared the dev branch against `upstream/main = c359ae6` on `.github/`, `CLAUDE.md`, `.github/copilot-instructions.md`, every `_quarto*.yml`, and the `Validation-Report/` directory tree. The two found (CI workflow + `de/` po4a targets) were the only stale-base hunks; the rest of the diff is intentional migration content. The CI workflow regression was the most consequential class — would have failed on push.

**On the profile-files correction:** noted, no action needed beyond acknowledgement. The conclusion held (top-level `*-Report.<lang>.qmd` are *per-language entrypoint documents*, not po4a-generated; syncing `params:` is legitimate); the "Quarto profile files" label was wrong. Thanks for the correction.

CI on `df7a9dd` will exercise the restored po4a step against the real `po/*.po4a.cfg` paths — expect green now.

Brar:
## Round 2 — verified `df7a9dd`, with one incomplete fix

Confirmed landed:
- ✅ `build.yml` restored — pinned `tools/po4a` + `po/*.po4a.cfg` paths + `submodules: true` + the gettext/perl deps. CI po4a step should pass.
- ✅ `includeAntibioticUtilisationTable: false` added to all five localized `Reference-Report.{de,el,es,et,it}.qmd` (matching `Reference-Report.qmd:39`) — the localized-render crash is resolved.
- ✅ `Validation-Report/de/` fully reverted — `funktionert` typo gone, the five deleted `_solution_*.Rmd` restored, and `de/` is no longer in the diff.
- ✅ `en/_problem_detail_0014.Rmd` body typos fixed.
- ✅ Stale-base sweep done + `New-PartnerReports.ps1` alignment filed as a task. 👍

**One incomplete item — the `ICHE` → `ICHI` rename missed several spots, including user-facing strings.** You fixed the body of `_problem_detail_0014.Rmd`, but `ICHE` still remains at:
- `reports/Validation-Report/en/_problem_detail_0014.Rmd:1` — the **section heading** ("…Invalid **ICHE** Code"), which renders while the body now says ICHI.
- `reports/Validation-Report/content/_sR.yaml:124,128,132` — the **validation error messages shown to partners** ("The surgical procedure ('%s') has an invalid **ICHE** code ('%s')…"). These are the most consequential.
- `reports/Validation-Report/_mapping.qmd:62` — a comment (cosmetic).

Since you've opened the rename window on this term, worth completing it in the same pass (per the "don't leave rename-window inconsistencies" guardrail). The `_sR.yaml` English-source edits will need `Invoke-Localization.ps1 -Update` to refresh the `msgid`s — which fits the PR-12-translations split you've already planned.

---

Brar:
## `build-docs` CI failure — root cause (introduced by this PR)

The failing [`build-docs` run](https://github.com/NeoIPC/Surveillance-Toolkit/actions/runs/26907206775/job/79375241757) dies with hundreds of:
```
asciidoctor: WARNING: image to embed not found or not readable: .../doc/protocol/img/AWaRe-W.svg   (also -A, -R)
##[error]Process completed with exit code 1
```

**Cause:** this PR moves `AWaRe-{A,W,R}.svg` out from under the protocol — `doc/protocol/img/ → reports/common/img/` (for the reports' new antibiotic-utilisation table). But the **protocol** still embeds them: `NeoIPC-BuildTools.psm1:421` emits `image:AWaRe-<category>.svg`, and the protocol header sets `:imagesdir: ./img` → `doc/protocol/img/`. After the move, that directory no longer has them.

`Make-NeoIPC-Core-Protocol.ps1` was updated to *source* them from the new location (`$commonImgDir`, L33/L98/L103/L108 ✅) but the copy **destination is still `$artifactsFolder/img/`** (L99/L104/L109) — nothing places them into `doc/protocol/img/`, which is asciidoctor's `imagesdir`. At base this never mattered because the SVGs were committed directly in `doc/protocol/img/`; the move removed asciidoctor's only source. The CI Build step runs with `-WarningAction:Stop`, so the resulting "image not found" warnings are fatal → exit 1.

**Fix options:**
1. Copy the three SVGs from `reports/common/img/` into **`$imgDir`** (`doc/protocol/img/`) during the build — i.e. change the AWaRe `*Dest` to `$imgDir` (matching where the XSLT-generated SVGs already land, L145/150/…), instead of (or in addition to) `$artifactsFolder/img/`. That puts them back under asciidoctor's `imagesdir`.
2. Or keep canonical copies committed in `doc/protocol/img/` and copy *out* to `reports/common/img/` for the reports — the reverse direction.

Option 1 is the smaller change and keeps a single source of truth in `reports/common/img/`. This is a genuine cross-consumer coupling (the SVGs feed both the protocol PDF and the reports), so a quick local `Make-NeoIPC-Core-Protocol.ps1` run is worth doing before the next push to confirm green — the `-WarningAction:Stop` strictness means any remaining missing-image warning will fail it.

---

Brar:
## Re: where the shared SVGs should live — agreed, `common/img/` (top-level)

Looked at the asset layout, and the maintainer's call is the right one. These AWaRe SVGs are **cross-consumer** (the protocol PDF *and* the reports' antibiotic-utilisation table), so putting them under `reports/common/img/` is the wrong frame — it makes the protocol build reach into a *reports* subdirectory (`$commonImgDir = …/reports/common/img`), which is what coupled them and broke `build-docs`.

The repo already has a top-level `common/` for shared assets (`common/logos/`), so the SVGs belong at **`common/img/`**, next to it. And there's a real duplication problem that supports unifying further:

- `common/logos/` — **referenced nowhere** (orphaned).
- `doc/protocol/img/LOGO_NEOIPC*.png` — the copy the protocol actually uses (`:title-logo-image: image:LOGO_NEOIPC.png`).
- `reports/logos/LOGO_NEOIPC*.png` (+ `eu-flag*`) — the copy the reports use.

So the logos are triplicated with the "shared" one unused. A single **`common/img/`** holding the shared images (logos + EU flag + AWaRe) as the one source of truth, with each build copying what it needs into its own `imagesdir`, removes that — your "just have `img` and move the logos there" option.

**Concretely for this PR:**
1. Move `AWaRe-{A,W,R}.svg` → `common/img/` (top-level), and point `Make-NeoIPC-Core-Protocol.ps1`'s `$commonImgDir` at `$workspaceFolder/common/img`.
2. **Fix the copy destination** — the SVGs must land in `doc/protocol/img/` (asciidoctor's `:imagesdir: ./img`), i.e. AWaRe `*Dest = $imgDir`, not `$artifactsFolder/img`. That's the part that actually unbreaks `build-docs`.
3. Reports reference them from `common/img/` (or a copy step into the reports' asset dir, mirroring how `reports/logos/` works today).

The full logo de-duplication is a larger cleanup and could be its own follow-up, but relocating AWaRe to `common/img/` + fixing the copy-dest is in-scope here since this PR introduced both the misplacement and the break. Worth a local `Make-NeoIPC-Core-Protocol.ps1` run to confirm green (the `-WarningAction:Stop` build fails on any remaining missing-image warning).

---

Brar:
**Correction to my previous comment** — I was wrong to call `common/logos/` "orphaned," thanks for the catch.

The protocol title page **does** use those images: `doc/protocol/xslt/NeoIPC-Core-Title-Page.xslt:77-78` references `cc.xlarge.png` / `by.xlarge.png` (the CC-licence badges), and the header references `LOGO_NEOIPC.png`. What's actually true:

- `cc.xlarge.png` / `by.xlarge.png` / `LOGO_NEOIPC*.png` are committed in **both** `common/logos/` **and** `doc/protocol/img/`.
- The title-page SVG references them by **bare filename**, so they resolve relative to the generated SVG's location (`doc/protocol/img/`) — the build consumes the `doc/protocol/img/` copies.
- No current file references the `common/logos/` *path*, and no build step copies from it — so `common/logos/` is an **un-wired parallel copy**, not an unused asset. The images themselves are in use (from the duplicated `doc/protocol/img/` and `reports/logos/` locations).

I conflated "the `common/logos/` path isn't referenced" with "these images are unused" — only the former is true. The substance of the consolidation argument is unchanged (and arguably stronger): `LOGO_NEOIPC.png` + the CC badges are duplicated across `common/logos/`, `doc/protocol/img/`, and `reports/logos/`, with `common/logos/` sitting un-wired — exactly the drift risk a single `common/img/` source-of-truth (with a build-time copy into each consumer's `imagesdir`) would remove. That's the established shared-asset pattern the AWaRe SVGs should follow too.

---

Brar:
## Round 2 responses

**ICHE → ICHI rename completed.** \`5f50161\` fixes the section heading (\`en/_problem_detail_0014.Rmd:1\`), the three user-facing error messages (\`content/_sR.yaml:124/128/132\`), and the mapping comment (\`_mapping.qmd:62\`). The German \`de/_problem_detail_0014.Rmd\` references stay as-is — that file is po4a-generated and will be refreshed via \`Invoke-Localization.ps1 -Update\` in PR-12-translations once these English source changes flow into the \`.pot\` and then the \`.po\`.

**\`build-docs\` CI failure — AWaRe SVGs relocated to top-level \`common/img/\`.** Three findings while implementing this:

1. The **reports don't reference the SVG file paths at all** — \`tables/_tbl-antibiotic-utilisation.qmd\` uses inline HTML badges via \`aware_colors\` (lines 47-69), not the SVG files. So the dev branch's move to \`reports/common/img/\` wasn't even serving the reports; it just broke the protocol. The only consumer is the protocol via \`image:AWaRe-<category>.svg\` in \`NeoIPC-BuildTools.psm1:421\`.
2. Done in \`5f50161\`: \`AWaRe-{A,W,R}.svg\` moved to top-level \`common/img/\`; \`Make-NeoIPC-Core-Protocol.ps1\`'s \`\$commonImgDir\` repointed at \`workspaceFolder/common/img\`; **\`AWaReA/W/RDest\` changed from \`\$artifactsFolder/img/\` to \`\$imgDir\`** — that last part is what was actually broken (\$imgDir is asciidoctor's \`:imagesdir: ./img\`).
3. **\`\$artifactsFolder/img/\` is the wrong destination across all three AWaRe copies, not just the dev-branch move artefact.** This was already wrong on the dev branch — asciidoctor never read from \`\$artifactsFolder/img\`. The AWaRe copy step only worked at base because the SVGs were committed directly to \`doc/protocol/img/\` and asciidoctor read those rather than the copies. So the dev-branch move *exposed* a latent dead-copy bug in the copy step rather than creating the asymmetry.

**Logo / shared-asset consolidation filed as follow-up.** [\`tasks/shared-assets-consolidation.md\`](tasks/shared-assets-consolidation.md) tracks the broader cleanup (migrate \`common/logos/\` → \`common/img/\`, delete duplicated LOGO_NEOIPC.png + CC badges from \`doc/protocol/img/\` and \`reports/logos/\`, update build steps to copy from the single source of truth). Out of scope here but the \`common/img/\` pattern this PR establishes is the foundation. Re: the orphan-vs-un-wired-parallel-copy correction on \`common/logos/\` — noted, the follow-up task description matches your refined characterization.

**Re: po4a-generated files (user question raised mid-pass).** Sanity-checked the staged tree: only \`content/_sR.yaml\` is staged (English source — committed by design); the 14 added \`.po\` files from the dev branch stay unstaged for PR-12-translations to pick up via \`Invoke-Localization.ps1 -Update\`. The de/ \`_problem_detail_*.Rmd\` files reverted in round 1 stay reverted. My earlier "translation files in staged" grep used \`_sR\.yaml\$\` which would have wrongly matched the English source — fixed in my head, not in code; the staged tree is correct.

CI \`build-docs\` on \`5f50161\` will exercise the \`\$imgDir\` copy destination — expect green now. I won't be able to run \`Make-NeoIPC-Core-Protocol.ps1\` locally before this push (the .NET XSLT-pipeline requires the full asciidoctor toolchain), so CI is the verification.

Brar:
## Round 2 — verified `5f50161`, and `build-docs` is green ✅

Confirmed the [`build-docs` run on `5f50161`](https://github.com/NeoIPC/Surveillance-Toolkit/actions/runs/26908611401/job/79380246255) is **success** (vs the failure on `df7a9dd`). All items check out:

- ✅ **`build-docs` fixed** — `AWaRe-{A,W,R}.svg` relocated to top-level `common/img/`; `$commonImgDir` → `…/common/img`; and the decisive change, AWaRe `*Dest` → `$imgDir` (`doc/protocol/img/`, asciidoctor's `:imagesdir`). CI green confirms it.
- ✅ **ICHE → ICHI** complete — heading, the three `_sR.yaml` error messages, and the `_mapping.qmd` comment are all clean now.
- ✅ Verified your finding that the reports use **inline HTML badges** (`aware_colors` + `<span style="…background:…">`), not the SVG files — no report references `AWaRe-*.svg` or `common/img`. So the SVGs are protocol-only and the original `reports/common/img/` move was pure breakage; good catch that `$artifactsFolder/img` was a pre-existing dead-copy too.

Nice work — that closes everything I'd flagged on this PR. The shared-asset consolidation (de-duplicating `LOGO_NEOIPC.png` + CC badges across `common/logos/`, `doc/protocol/img/`, `reports/logos/` into the `common/img/` source-of-truth this PR establishes) is rightly a follow-up. No open items from me.

---

PR #72, opened 2026-06-15
Report-layer source of truth: presets, profile cleanup, locale wrappers

Brar:
Report-layer half of the cross-repo report-form design pass — makes the report directory the single source of truth for report configuration.

- Adds per-report `presets.json` (Default / Minimal / Full) that the backend serves at runtime and the app applies as `includeX` parameters.
- Removes the non-functional top-level `_quarto-{default,full}.yml` profiles and the `[full, default]` group; the render profile is derived from locale + output format.
- Removes the Reference report's five per-section text parameters — text inclusion is governed solely by `includeIntroductionTexts` / `includeMethodsTexts`, matching the Partner report.
- Adds `Build-LocaleReportSources.ps1` and regenerates the per-locale report wrappers.


Brar (commented):
## Review — Report-layer source of truth: presets, profile cleanup (#72)

Clean PR. I verified the risk areas directly:

**✅ Verified solid**
- **Localized param-set sync** (the recurring drift risk) — after removing the five `includeText*` section params, the master `Reference-Report.qmd` and all five `Reference-Report.{de,el,es,et,it}.qmd` variants have **identical** param key sets; `includeIntroductionTexts`/`includeMethodsTexts` retained; no `includeText*` token survives anywhere.
- **`_content.qmd` / conditional refs** — no dangling `params$includeText*`; the patient-population heading became unconditional but is safe (no `@sec-`/`@tbl-`/`@fig-` cross-reference targets it), and section visibility is now governed by `.show(params$includeIntroductionTexts/MethodsTexts)`. (Four of the five removed params were already dead in the base — never referenced — so this is largely dead-param cleanup.)
- **`presets.json`** (both reports) — well-formed; every preset key maps to a real `params:` entry; `includeConfidenceIntervals` values stay within the `@values [all, rate, none]` enum; the Partner `full`→`"all"` vs Reference (default already `"all"`) asymmetry is intentional and correct.
- **Profile removal** — `_quarto-{default,full}.yml` + the `[full, default]` group are gone with no dangling refs (the retained `_quarto-minimal.yml` legitimately sets a real `format` option, unlike the removed ones which couldn't set document params).
- **Scripts** — `New-ReferenceReport.ps1`'s `$elementMapping` correctly drops all five `includeText*` keys (intro/methods now gated globally via `-HideIntroductionTexts`/`-HideMethodsTexts`); `Build-LocaleReportSources.ps1` is well-formed (`Build` is an approved PS verb, `[CmdletBinding()]`, `-Check` mode), and its transform reproduces the committed wrappers.

**Cross-repo contract — aligned.** The 16 `include*` flags here match NeoIPC-Reporting #15's schema JSON (regenerated from this PR's `b3f6ddf`) and app #10's consumption; the removed `includeText*` params are gone from the backend schema too.

**🟢 Two minor / non-blocking:**
1. **Stale `.gitignore` entries** — `reports/.gitignore:4-5` still has `!_quarto-default.yml` / `!_quarto-full.yml` whitelisting the now-deleted profiles. Harmless (they re-include nonexistent files) but worth removing in this pass for cleanliness. (The PR doesn't currently touch `.gitignore`.)
2. **Description nit** — the "Also carries an instruction-file guardrail sync (no force-push under open review)" line doesn't match the diff: this PR touches neither `CLAUDE.md` nor `.github/copilot-instructions.md`; that guardrail already landed in the base via #71. Worth trimming from the description/squash message.

Nothing blocking. (Comment, not approval — GitHub blocks approving your own PR.)

---

Brar:
Both addressed.

**1. Stale `.gitignore`** — fixed in `8af160d`: dropped the `!_quarto-default.yml` / `!_quarto-full.yml` whitelist entries for the profiles deleted in this PR.

**2. Instruction-file guardrail** — one correction: the PR *does* touch `CLAUDE.md` and `.github/copilot-instructions.md` (commit `dd55621`), so "touches neither" isn't quite right. But the underlying point holds — the no-force-push guardrail already landed in base via #71 (confirmed present in both files on `upstream/main`), so `dd55621` is a redundant add/add. The PR is mergeable (git dedupes it — no duplicate, no conflict), so I left the commit as-is (removing it would mean force-pushing an open PR, which is exactly what that guardrail forbids) and trimmed the over-claiming "instruction-file guardrail sync" line from the description instead.

Brar:
Both verified — and you're right on the correction, conceding it.

1. **`.gitignore`** — `8af160d` drops the `!_quarto-default.yml` / `!_quarto-full.yml` entries. ✅
2. **Instruction files** — my "touches neither" was imprecise. The PR *does* carry a commit (`dd55621`) that modifies `CLAUDE.md` + `.github/copilot-instructions.md`; the net `base..head` diff is empty only because #71 already added the identical line (add/add dedup) — which is what my `git diff --name-only` was reporting. So the *net effect* is nil, but the PR isn't "untouched." Trimming the description line (rather than the commit) is the right call — and the can't-drop-it-without-force-pushing-an-open-PR point is a neat self-demonstration of the very guardrail. 👍

Nothing else from me on this one.

---

PR #73, opened 2026-06-19
Resolve report -OutputDir against PowerShell's location, not .NET's CWD

Brar:
The report wrapper scripts (`scripts/New-*.ps1`) accept a `-OutputDir`. A relative one was resolved with .NET path APIs that use `[Environment]::CurrentDirectory`, which PowerShell does not keep in sync with its own location (`$PWD`). So a relative `-OutputDir` (e.g. `./data`) could resolve against the wrong base — landing files outside the intended directory — and the build-report write then failed on the missing directory because `Set-Content` does not create missing parents.

The four non-Reference wrappers avoided the wrong-base half, but — declaring `SupportsShouldProcess` — carried a latent `-WhatIf` failure: a suppressed `New-Item` followed by a `Resolve-Path` that throws on the not-yet-created directory.

All five wrappers now resolve `-OutputDir` against `$PWD` and create the directory up front, which keeps `-WhatIf` dry runs side-effect-free, with a fallback for the exotic case of a caller sitting on a non-filesystem PSDrive.

### DIFF
--- a/scripts/New-PartnerCertificate.ps1
+++ b/scripts/New-PartnerCertificate.ps1
@@ -84,7 +84,7 @@ if ($OutputDir) {
     # PSDrive). GetFullPath creates nothing (stays -WhatIf-safe); New-Item makes the dir.
     $base = if ([System.IO.Path]::IsPathFullyQualified($PWD.ProviderPath)) { $PWD.ProviderPath } else { [Environment]::CurrentDirectory }
     $outputDirPath = [System.IO.Path]::GetFullPath($OutputDir, $base)
-    if (-not (Test-Path -LiteralPath $outputDirPath)) {
+    if (-not (Test-Path -LiteralPath $outputDirPath -PathType Container)) {
         New-Item -ItemType Directory -Path $outputDirPath -Force | Out-Null
     }
     $outputDirExplicit = $true
--- a/scripts/New-PartnerCertificate.ps1
+++ b/scripts/New-PartnerCertificate.ps1
@@ -78,12 +78,15 @@ $reportDirPath = Resolve-Path -LiteralPath "$PSScriptRoot/../reports/Partner-Cer
 
 # Resolve OutputDir BEFORE changing directory (it's relative to the caller's CWD)
 if ($OutputDir) {
-    $outputDirPath = Resolve-Path -LiteralPath $OutputDir -ErrorAction SilentlyContinue
-    if (-not $outputDirPath) {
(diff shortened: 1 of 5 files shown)


Brar (commented):
## Review — Resolve report `-OutputDir` against PowerShell's location, not .NET's CWD

Correct, well-reasoned hotfix. Static review (pwsh isn't installed here), but the .NET/PowerShell semantics are well-defined and the three-part fix holds:

- **Wrong base (the headline)** — `[System.IO.Path]::GetFullPath($OutputDir, $base)` with `$base = $PWD.ProviderPath` resolves a relative `-OutputDir` against PowerShell's *location* instead of `[Environment]::CurrentDirectory`. That's the right knob: `$PWD.ProviderPath` is the filesystem-native current location, and the 2-arg `GetFullPath` ignores `$base` for an already-absolute path. ✅
- **`-WhatIf` safety** — confirmed all five scripts declare `SupportsShouldProcess`, so the inner `New-Item` inherits `$WhatIfPreference` and is suppressed on a dry run (no directory created → side-effect-free). And because the path is now computed with `GetFullPath` (pure path math, no filesystem touch) rather than the old `Resolve-Path`, it no longer throws on the not-yet-created directory. That precisely closes the latent `-WhatIf` failure the four non-Reference wrappers carried. ✅
- **Missing-dir-on-write** — `New-Item -ItemType Directory -Force` creates the directory (and any parents) up front in the real run, so the later `Set-Content`/`ConvertTo-Json` build-report write no longer fails on a missing parent. This also fixes the Reference wrapper, whose old "directory created on first write" comment was wrong (`Set-Content` doesn't create parents). ✅
- **Non-filesystem PSDrive fallback** — `IsPathFullyQualified($PWD.ProviderPath)` → `[Environment]::CurrentDirectory` is a sensible best-effort for the exotic case of a caller sitting on a Cert:/HKLM:/custom PSDrive.

The fix is applied **identically** across all five wrappers (good — no per-script drift), and no `Co-Authored-By` trailer. The 2-arg `GetFullPath` and `IsPathFullyQualified` are .NET Core+ APIs, which is consistent with the repo's existing PS7 requirement (`Get-Date -AsUTC`, etc.).

🟢 One optional robustness nit (non-blocking): `Test-Path -LiteralPath $outputDirPath` returns true for *any* existing item, so if `-OutputDir` happened to resolve onto an existing **file**, `New-Item` is skipped and the later write fails. `-PathType Container` would make the guard precise. Exotic misuse, so your call.

LGTM. (Comment, not approval — GitHub blocks approving your own PR.)

---

PR #74, opened 2026-06-24
CSV/YAML to DHIS2 metadata pipeline: canonical metadata directory + installable package

Brar:
Makes a version-controlled `metadata/` directory the canonical, reviewable, diffable, translatable source of NeoIPC's DHIS2 metadata, and generates the importable package from it. The conversion is pure file processing — no DHIS2 API calls (it operates on an out-of-band, PII-cleaned export on disk).

## What it delivers

- **Round-trippable per-type converter** — a DHIS2 metadata export ↔ a directory of one file per object type, in the most-suitable format per shape (CSV for tabular types, YAML for nested ontology, named `sharing` profiles, one text file per DHIS2 expression), with semantic round-trip equality modulo normalisation.
- **Node-free dependency closure** seeded at the `NEOIPC_CORE` program (DHIS2's own dependency export silently drops expression-embedded references), plus a NeoIPC-specific expression linter and a canonicaliser for the name-argument d2-functions.
- **Full coverage** — the org-unit family and access-control objects, and the matrix-generated families (per-slot pathogen/substance data elements; resistance / field-gating / substance rules, variables and actions) materialised as ordinary directory rows; the option domain (pathogens from the infectious-agents YAML, antibiotics from the curation CSVs) generated at build.
- **Export-independent build** — `New-NeoIPCMetadataPackage` assembles the importable package from the directory alone. The directory models the corrected *to-be* state, not the flawed deployed export (it carries the resistance-description typo fix, the BSI no-positive-culture rule promotion, and drops a retired aggregate rule and a legacy duplicate birth-weight data element).
- **gettext-PO translations** for the metadata and antibiotic catalogues, keyed by a stable semantic context so a rename or slot addition stays change-local.
- **Reconcile mode** to classify drift between a deployed instance and the directory.
- **Installable alpha packages** under `metadata/dist/` (an install base, and a play package with a synthetic test hierarchy and users) so NeoIPC can be installed without running the pipeline. Each carries a minimal top-level `package` manifest, which the importer skips as an unknown schema.

## Deferred to separate tasks

- A standards/WHO-compatible package, and the user-group / role / permission model it depends on.
- Migrating the production instance to the canonical metadata and removing the remaining seed-bridging code.
- Registering the new PO components in Weblate (post-merge).

Surveillance-Toolkit runs standalone; all changes are under `scripts/` and `metadata/`. The NeoIPC-Tools Pester suite is green.


scripts/modules/NeoIPC-Tools/Private/MetadataAssembly.ps1
```diff
@@ -0,0 +1,395 @@
...
+    # Salvage hand-authored actions (target DE outside the generated families) from each replaced rule
+    # onto the generated rule of the same id (which carries the deployed id where reproduced).
+    $salvagedActions = [System.Collections.Generic.List[object]]::new()
+    foreach ($r in @($Config['programRules'])) {
+        if ($r -isnot [System.Collections.IDictionary]) { continue }
+        $rid = [string]$r['id']
+        if (-not ($genRuleIds.Contains($rid) -or $genRuleNames.Contains([string]$r['name']))) { continue }
+        $genRule = $genRuleById[$rid]
+        if (-not $genRule) { continue }   # replaced by name but its id was re-minted: no same-id gen rule to attach to
+        foreach ($a in @($deployedActionsByRuleId[$rid])) {
```
Brar, line 358: **Latent crash: unguarded index of a possibly-`$null` action.**

When `$rid` is absent from `$deployedActionsByRuleId`, `$deployedActionsByRuleId[$rid]` is `$null` and `@($null)` is a **1-element array `[$null]`** (not empty), so the loop body runs once with `$a = $null`. The very next line — `$genActionIds.Contains([string]$a['id'])` — then throws `Cannot index into a null array` and aborts the whole package assembly.

Trigger: a generated-class program rule that is present in the export but referenced by **zero** `programRuleActions` (a condition-only rule, or one whose actions were deleted in DHIS2). Not hit by the current canonical export (every matrix rule has actions today), so this is latent rather than live — but it's the one `@(...)` loop in the module that indexes the element *before* an `-is [IDictionary]` guard.

Suggested fix — skip rules with no deployed actions:
```powershell
foreach ($a in @($deployedActionsByRuleId[$rid])) {
    if ($a -isnot [System.Collections.IDictionary]) { continue }
    ...
```
or guard the lookup with `if (-not $deployedActionsByRuleId.ContainsKey($rid)) { continue }` before the inner loop.

---

scripts/modules/NeoIPC-Tools/Private/MetadataTypeMaps.ps1
```diff
@@ -0,0 +1,346 @@
...
+        trackedEntityTypeAttribute = 'bool'; userAttribute = 'bool'; userGroupAttribute = 'bool'
+        validationRuleAttribute = 'bool'; validationRuleGroupAttribute = 'bool'; visualizationAttribute = 'bool' } }
+    'organisationUnits'     = @{ NaturalKey = 'code'; Nesting = 'TopLevel'; Properties = [ordered]@{
+        code = 'string'; name = 'string'; shortName = 'string'; openingDate = 'string'; closedDate = 'string'; level = 'int'; parent = 'id'; image = 'id' } }
+    'organisationUnitGroups' = @{ NaturalKey = 'code'; Nesting = 'TopLevel'; Properties = [ordered]@{
+        code = 'string'; name = 'string'; shortName = 'string'; description = 'string'; symbol = 'string'; color = 'string' } }   # organisationUnits (membership) is per-deployment — stripped on capture; re-authored as common + play overlays
+    'organisationUnitGroupSets' = @{ NaturalKey = 'code'; Nesting = 'TopLevel'; Properties = [ordered]@{
+        code = 'string'; name = 'string'; shortName = 'string'; description = 'string'
+        compulsory = 'bool'; dataDimension = 'bool'; includeSubhierarchyInAnalytics = 'bool'; organisationUnitGroups = 'idArray' } }
```
Brar: **Possible round-trip reordering — please confirm against the hbm.**

`organisationUnitGroupSets.organisationUnitGroups` is typed `'idArray'`, so the normalizer ordinal-sorts it and the authored member-group order is **not** preserved on round-trip / assembly. But `optionGroupSets.optionGroups` (line 148) — the structurally-analogous group-set→group `<list>` — *is* typed `'idArrayOrdered'`.

If DHIS2 maps `OrganisationUnitGroupSet.organisationUnitGroups` as an ordered `<list>` (the way it does `OptionGroupSet.optionGroups`), then a group-set whose member groups are authored in a non-alphabetical order will silently come back reordered, changing the analytics dimension item order. The file comment says the `idArrayOrdered` set was hbm-verified, so this is either a deliberate set-mapping or a miss — worth a 30-second check against `refs/dhis2-core` `OrganisationUnitGroupSet.hbm.xml` before merge. (I couldn't verify it in the review environment — no `refs/` checkout.)

---

scripts/modules/NeoIPC-Tools/Private/MetadataTranslation.ps1
```diff
@@ -0,0 +1,656 @@
...
+}
+
+function Test-NeoIPCAtcCode {
+    # True when $Code is a WHO ATC code at the antibiotic-relevant levels: ATC level 4 (5 chars, e.g. J01CG — the
+    # antibiotic option GROUPS) or level 5 (7 chars, e.g. J01AA01 — the substance OPTIONS). A pure code-shape test;
+    # callers decide what to do with the answer.
+    [CmdletBinding()]
+    [OutputType([bool])]
+    param([AllowEmptyString()][AllowNull()][string]$Code)
+    return $Code -cmatch '^[A-Z][0-9]{2}[A-Z]{2}([0-9]{2})?$'
```
Brar, line 637: **Altitude: the ATC-4 code-shape predicate is inlined at four sites, and this one has drifted.**

The "is this an antibiotic-domain ATC code" test is hand-written in four places, and they are **not identical**:

- `Private/MetadataTranslation.ps1:635` (this `Test-NeoIPCAtcCode`) — `^[A-Z][0-9]{2}[A-Z]{2}([0-9]{2})?$` (matches ATC-4 **and** 7-char ATC-5)
- `Private/Metadata.ps1:747` (`Test-NeoIPCMetadataGeneratedExcluded`) — `^[A-Z][0-9]{2}[A-Z]{2}$` (ATC-4 only)
- `Public/AntibioticGeneration.ps1:331` — `^[A-Z][0-9]{2}[A-Z]{2}$` (ATC-4 only)
- `Public/GenerationDiff.ps1:102` — `^[A-Z][0-9]{2}[A-Z]{2}$` (ATC-4 only)

The emit, comparator, classified-diff and translation paths each decide "is this object antibiotic-domain" from their own copy. Harmless today (option groups are always ATC-4, so the optional `([0-9]{2})?` never fires), but it's exactly the "definitions live ONCE" invariant this module is built on — a future ATC-shape change updates some-but-not-all and the four paths silently disagree on what to exclude. Recommend exporting this `Test-NeoIPCAtcCode` helper (one canonical regex) and calling it from the other three sites.

---

scripts/modules/NeoIPC-Tools/Private/MetadataAuthoring.ps1
```diff
@@ -0,0 +1,387 @@
...
+    param([Parameter(Mandatory)][string[]]$Path)
+
+    $objects = [System.Collections.Generic.List[System.Collections.Specialized.OrderedDictionary]]::new()
+    # code and id are both unique identities across all files (common scaffold + variant overlay). A repeated
+    # code would silently clobber membership / user-assignment resolution (those key on code, last-write-wins);
+    # a repeated id collides at idScheme=UID import. A malformed/blank id would fail only at import — and a
+    # blank id slips past the assembly collision guard on first sight — so reject ids up front, here, against
+    # the offending code (symmetric with the user path's Test-NeoIPCMetadataUid).
+    $seenCode = @{}
+    $seenId = @{}
```
Brar: **Case-sensitivity inconsistency: UID dedup is case-insensitive, but DHIS2 UIDs are case-sensitive.**

`$seenId = @{}` is a case-**insensitive** hashtable, yet DHIS2 UIDs are case-**sensitive** (11 chars, mixed case) and the downstream assembly collision guard (`Join-NeoIPCMetadataPackage`) uses `[StringComparer]::Ordinal`. So the two checks disagree on what a collision is: two legitimately-distinct authored UIDs differing only in case would be wrongly rejected here as a "Duplicate authored org-unit UID" even though the Ordinal assembly check would accept them.

Low probability (UIDs are random), and it errs safe (false rejection, not silent acceptance), but for consistency `$seenId` should be an ordinal-cased map, e.g.:
```powershell
$seenId = [System.Collections.Generic.HashSet[string]]::new([System.StringComparer]::Ordinal)
# ... if (-not $seenId.Add($id)) { throw ... }
```
`$seenCode` is arguably the same (DHIS2 codes are case-sensitive). The username `$seen` map (~line 141) can stay case-insensitive if DHIS2 usernames are case-folded.

---

scripts/modules/NeoIPC-Tools/Private/MetadataExpression.ps1
```diff
@@ -0,0 +1,330 @@
...
+    # hashtable: Package (the transformed clone) + per-transform counts (+ UidMap when regenerating).
+    [CmdletBinding()]
+    [OutputType([hashtable])]
+    param(
+        [Parameter(Mandatory)]$Package,
+        [switch]$Canonicalize,
+        [switch]$RegenerateUids
+    )
+    # Deep clone via JSON round-trip (the package is JSON-origin; PS 7.5 -AsHashtable preserves key order).
+    $clone = ConvertFrom-NeoIPCMetadataJsonText -Json ($Package | ConvertTo-Json -Depth 100)
```
Brar, line 310: **Efficiency: the whole package is deep-cloned even when nothing (or only `-Canonicalize`) needs it.**

`Update-NeoIPCMetadataPackage` always serializes the entire package to JSON at depth 100 and re-parses it, before checking which switches were passed. Called from `Public/Metadata.ps1:346` with whatever the user supplied — so a `-Canonicalize`-only call (which mutates a handful of program-rule/action condition strings) or a no-op call pays a full multi-megabyte serialize+parse round-trip over the 3500+-option canonical package.

The full-tree clone is only needed for `-RegenerateUids` (which rewrites tokens everywhere). For canonicalize-only, the slots are addressed via `Get-NeoIPCMetadataExpressionSlot` and could be cloned/replaced in place; and with neither switch set the input could be returned unchanged. Not a hot path (one-shot CLI op), but an easy win.

---

scripts/modules/NeoIPC-Tools/Private/MetadataClosure.ps1
```diff
@@ -0,0 +1,361 @@
...
+    [OutputType([System.Collections.Specialized.OrderedDictionary])]
+    param(
+        [Parameter(Mandatory)]$Base,
+        [Parameter(Mandatory)]$Supplement,
+        [string[]]$Types = @('programNotificationTemplates')
+    )
+    $merged = [ordered]@{}
+    foreach ($k in $Base.Keys) { $merged[$k] = $Base[$k] }
+    foreach ($t in $Types) {
+        if ($Supplement -is [System.Collections.IDictionary] -and $Supplement.Contains($t) -and @($Supplement[$t]).Count -gt 0) {
```
Brar: **Minor: the "has content" guard is truthy for a present-but-`$null` value.**

`@($Supplement[$t]).Count -gt 0` is `1 > 0` (true) when `$Supplement[$t]` is `$null`, because `@($null)` is a 1-element array. So a supplement that carries the key with a `$null` value merges `$null` into `$merged[$t]` and skips the `Write-Warning`, rather than warning that there's nothing to splice. Realistic supplements either omit the key or carry a real array, so impact is low — but the guard doesn't do what it reads as. Test the value explicitly, e.g. `$Supplement[$t] -and @($Supplement[$t]).Count -gt 0`.

---

Brar (commented):
## Code review — CSV/YAML→DHIS2 metadata pipeline

Reviewed the ~13k lines of new/changed PowerShell (the 700+ generated `.dhis2`/`.csv`/`.po` data files were treated as out of scope for logic review). This is unusually high-quality, defensively-written code — the single-element-array-unrolling trap is handled with the unary-comma idiom throughout, `@(...)` loops are guarded with `-is [IDictionary]`, sorts are ordinal for cross-machine determinism, and invariants are documented inline. Most candidate issues from the first pass were **refuted** on verification (e.g. `Get-NeoIPCMetadataScopedConfig` does return its value because `Remove-NeoIPCMetadataNoise` returns `$Object`; the action-salvage loop accumulates correctly; the diff pairs by unique `id`; UID minting is culture-invariant).

Six findings survived, posted inline:

| # | Location | Kind | Severity |
|---|----------|------|----------|
| 1 | `MetadataAssembly.ps1:358` | Latent crash — unguarded index of a `$null` action (`@($null)` is a 1-element array) when a generated rule has no deployed actions | Med (latent; not hit by current data) |
| 2 | `MetadataTypeMaps.ps1:235` | `organisationUnitGroupSets.organisationUnitGroups` typed `idArray` vs the analogous `optionGroupSets.optionGroups` `idArrayOrdered` — possible silent member-order reordering on round-trip | Med (needs hbm confirm) |
| 3 | `MetadataTranslation.ps1:635` (+ `Metadata.ps1:747`, `AntibioticGeneration.ps1:331`, `GenerationDiff.ps1:102`) | Altitude — ATC-4 shape predicate inlined at 4 sites, one drifted; should be one shared helper | Med (maintainability) |
| 4 | `MetadataAuthoring.ps1:38` | UID dedup uses a case-insensitive map while UIDs are case-sensitive and the assembly guard is `Ordinal` | Low |
| 5 | `MetadataExpression.ps1:306` | Efficiency — whole package deep-cloned via JSON even for `-Canonicalize`-only / no-op calls | Low |
| 6 | `MetadataClosure.ps1:355` | Minor — `@($Supplement[$t]).Count -gt 0` is truthy for a present-but-`$null` value | Low |

**Severity note:** none of these is a confirmed failure on the canonical NeoIPC export. #1 is the most concrete (the crash mechanism is verified) but is gated by the fact that every current matrix rule has actions; #2 is the highest-impact *if* real but I couldn't verify it here (no `refs/dhis2-core` checkout). The rest are robustness/maintainability/efficiency.

**Conventions (CLAUDE.md):** clean — exported cmdlets carry comment-based help, no `Co-Authored-By` trailers, no absolute paths, no placeholder stubs, no personal names in source, and `CLAUDE.md`/`.github/copilot-instructions.md` are in sync.

Not blocking — flagging #1 and #2 as the two worth resolving (or consciously dismissing) before merge.

---

Brar:
All six findings addressed in `6d9ad66` — each verified against source first. 446 Pester tests green; the committed distribution artifacts are byte-identical (every change is internal/robustness, with no effect on generated output).

| # | Location | Disposition |
|---|----------|-------------|
| 1 | `MetadataAssembly.ps1` salvage loop | **Fixed** — added an `-isnot [IDictionary]` skip at the top of the loop, so a generated rule with **no** deployed actions (`@($null)` → a one-element `[$null]` array) is skipped instead of indexed. |
| 2 | `MetadataTypeMaps.ps1` `organisationUnitGroups` | **Not a bug — current typing is correct** (confirmed against `refs/dhis2-core`). `OrganisationUnitGroupSet.hbm.xml` maps `organisationUnitGroups` as an unordered `<set>` (no `<list-index>`), whereas `OptionGroupSet` maps `members` as an ordered `<list>` with `<list-index column="sort_order">`. There is no authored member order to preserve, so `idArray` (ordinal sort for determinism) is right and `idArrayOrdered` would be wrong. Documented the `<set>`-vs-`<list>` distinction inline so it isn't re-raised. |
| 3 | ATC-4 predicate (4 sites) | **Fixed** — extracted `Test-NeoIPCAtcGroupCode` (ATC-4 only) as the single source and routed the three group-classification sites through it. The fourth site, `Test-NeoIPCAtcCode`, is intentionally ATC-4-**or**-5 (it spans option groups *and* substance options) and now documents that so it isn't narrowed. |
| 4 | `MetadataAuthoring.ps1` dedup | **Fixed** — `$seenId`/`$seenCode` are now `Ordinal` `HashSet`s, matching the `Ordinal` assembly collision guard. |
| 5 | `MetadataExpression.ps1` clone | **Fixed** — `Update-NeoIPCMetadataPackage` returns the input unchanged when neither switch is set, skipping the multi-MB clone. |
| 6 | `MetadataClosure.ps1` merge guard | **Fixed** — added an explicit `$Supplement[$t] -and` value check before the count test, so a present-but-`$null` value warns instead of merging `$null`. |

The refuted/verified items from the first pass (the value getters returning via `Remove-NeoIPCMetadataNoise`, the salvage accumulation, id-paired diffing, culture-invariant minting) matched my reading as well.

PR #77, opened 2026-06-26
Add the NeoIPC Core data-dictionary generator

Brar:
## NeoIPC Core data dictionary

A technology-agnostic data dictionary (codebook) for the NeoIPC Core surveillance program, auto-generated from the canonical metadata so it stays in step with future changes. It sits between the machine-readable DHIS2 metadata and the end-user protocol — readable by both epidemiologists and technical implementers.

`Export-NeoIPCDataDictionary` (NeoIPC-Tools) flattens the assembled metadata package into four sheets:

- **About** — program identity, counts, provenance
- **Variables** — patient attributes, per-event data elements, and the per-event dates
- **Code lists** — every option set in full (including the pathogen and antimicrobial lists)
- **Forms & dates** — each module/form and its event date

emitted as CSV plus a multi-tab `.xlsx`. DHIS2 UIDs are omitted (technology-specific); the project variable code is the stable identifier.

- **Release artifact, not source:** output goes to the gitignored `artifacts/` tree and is generated in CI (uploaded with the build). See `docs/data-dictionary.md`.
- **`.xlsx`** via the MIT DocumentFormat.OpenXml SDK, restored with `dotnet` (`dotnet publish` of `scripts/modules/NeoIPC-Tools/lib`); the CSV path needs no extra dependency.
- Module baseline raised to PowerShell 7.6 / .NET 10.

Generate locally: `./scripts/New-NeoIPCDataDictionary.ps1` (CSV + `.xlsx`) or `-Format Csv`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,12 +15,12 @@ jobs:
         # runtime the NeoIPC-Tools module and data-dictionary generator require.
         runs-on: ubuntu-26.04
         steps:
-            - uses: actions/checkout@v4
+            - uses: actions/checkout@v7
               with:
                 submodules: true
 
             - name: Set up .NET 10 (for the data-dictionary .xlsx writer)
-              uses: actions/setup-dotnet@v4
+              uses: actions/setup-dotnet@v5
               with:
                 dotnet-version: '10.0.x'
 
@@ -58,7 +58,7 @@ jobs:
                 Install-Module -Name Pester -MinimumVersion 5.5 -Force -Scope CurrentUser -SkipPublisherCheck
                 # powershell-yaml: the metadata assembler reads the infectious-agents / sharing YAML when
                 # building the package the data dictionary flattens.
-                Install-Module -Name powershell-yaml -Force -Scope CurrentUser -SkipPublisherCheck
+                Install-Module -Name powershell-yaml -MinimumVersion 0.4.2 -Force -Scope CurrentUser -SkipPublisherCheck
                 Invoke-Pester -Path scripts/modules/NeoIPC-Tools/Tests/DataDictionary.Tests.ps1 -CI
 
             - name: Generate data dictionary
@@ -69,7 +69,7 @@ jobs:
                 ./scripts/New-NeoIPCDataDictionary.ps1 -OutputDirectory ./artifacts/data-dictionary
 
             - name: Upload artifacts
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@v7
               with:
                 name: NeoIPC-Docs-Preview
                 path: ./artifacts/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
               shell: bash
               run: |
                 sudo apt-get update -qq
-                sudo apt-get install -qq pandoc librsvg2-bin bison flex libffi-dev libxml2-dev libgdk-pixbuf2.0-dev libcairo2-dev libpango1.0-dev libwebp-dev fonts-lyx fonts-noto-core gettext liblocale-gettext-perl libtext-wrapi18n-perl libsyntax-keyword-try-perl libtimedate-perl libyaml-tiny-perl libpod-parser-perl libmodule-build-perl
+                sudo apt-get install -qq pandoc librsvg2-bin bison flex libffi-dev libxml2-dev libgdk-pixbuf-2.0-dev libcairo2-dev libpango1.0-dev libwebp-dev fonts-lyx fonts-noto-core gettext liblocale-gettext-perl libtext-wrapi18n-perl libsyntax-keyword-try-perl libtimedate-perl libyaml-tiny-perl libpod-parser-perl libmodule-build-perl
                 sudo gem install mathematical:1.6.18 asciidoctor-pdf asciidoctor-bibtex asciidoctor-diagram asciimath asciidoctor-mathematical
                 sudo npm install -g @linthtml/linthtml
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,11 +11,18 @@ on:
 
 jobs:
     build-docs:
-        runs-on: ubuntu-latest
+        # Pinned to an explicit image for reproducibility; 26.04 provides the .NET 10 / PowerShell 7.6
+        # runtime the NeoIPC-Tools module and data-dictionary generator require.
+        runs-on: ubuntu-26.04
         steps:
             - uses: actions/checkout@v4
               with:
                 submodules: true
+
+            - name: Set up .NET 10 (for the data-dictionary .xlsx writer)
+              uses: actions/setup-dotnet@v4
+              with:
+                dotnet-version: '10.0.x'
 
             - name: Prepare environment
               shell: bash
@@ -38,6 +45,23 @@ jobs:
             - name: Build
               shell: pwsh
               run: ./scripts/Make-NeoIPC-Core-Protocol.ps1 -InformationAction Continue -WarningAction:Stop
+
+            - name: Provision and test NeoIPC-Tools
+              shell: pwsh
+              run: |
+                # Provision the DocumentFormat.OpenXml assembly the .xlsx writer loads, then run the module
+                # tests so the OpenXmlValidator-backed XLSX schema-validity gate runs in CI (it is skipped only
+                # when the assembly is absent).
+                dotnet publish scripts/modules/NeoIPC-Tools/lib -o scripts/modules/NeoIPC-Tools/lib/bin -c Release --nologo
+                Install-Module -Name Pester -MinimumVersion 5.5 -Force -Scope CurrentUser -SkipPublisherCheck
+                Invoke-Pester -Path scripts/modules/NeoIPC-Tools/Tests/DataDictionary.Tests.ps1 -CI
+
+            - name: Generate data dictionary
+              shell: pwsh
+              run: |
+                # Generate the CSV + .xlsx data dictionary into the uploaded artifacts. Assembled from the
+                # metadata/ directory alone — no DHIS2 API.
+                ./scripts/New-NeoIPCDataDictionary.ps1 -OutputDirectory ./artifacts/data-dictionary
 
             - name: Upload artifacts
               uses: actions/upload-artifact@v4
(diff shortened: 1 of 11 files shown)


.github/workflows/build.yml
```diff
@@ -11,18 +11,27 @@ on:
...
-                sudo gem install mathematical:1.6.18 asciidoctor-pdf asciidoctor-bibtex asciidoctor-diagram asciimath asciidoctor-mathematical
+                sudo apt-get install -qq pandoc librsvg2-bin bison flex libffi-dev libxml2-dev libgdk-pixbuf-2.0-dev libcairo2-dev libpango1.0-dev libwebp-dev fonts-lyx fonts-noto-core gettext liblocale-gettext-perl libtext-wrapi18n-perl libsyntax-keyword-try-perl libtimedate-perl libyaml-tiny-perl libpod-parser-perl libmodule-build-perl
+                # CMAKE_POLICY_VERSION_MINIMUM lets the gems' bundled mtex2MML (an old cmake_minimum_required)
+                # configure under the CMake 4 on Ubuntu 26.04; passed through sudo, which otherwise resets env.
+                sudo env CMAKE_POLICY_VERSION_MINIMUM=3.5 gem install mathematical:1.7.0 asciidoctor-pdf asciidoctor-bibtex asciidoctor-diagram asciimath asciidoctor-mathematical
```
Brar, line 34: **CI fixes reviewed — all four are correct.** I verified each against upstream rather than just reading the diff:

- `libgdk-pixbuf2.0-dev` → **`libgdk-pixbuf-2.0-dev`** (line 31): correct — that's the renamed dev package and it resolves on the 26.04 archive.
- **`mathematical:1.7.0`** (line 34): the version exists on RubyGems.
- **`sudo env CMAKE_POLICY_VERSION_MINIMUM=3.5 gem install …`** (line 34): the right idiom — `sudo`'s `env_reset` strips an inline `VAR=val` prefix, so wrapping with `env` is what actually propagates the policy override to mtex2MML's CMake-4 configure. Note it applies to every gem in the command, which is harmless (only the mtex2MML build reads it).
- **`Install-Module powershell-yaml`** (line 61): genuinely required — the *generate* step calls `New-NeoIPCMetadataPackage`, which reads the infectious-agents / sharing YAML. (The Pester step above it doesn't need it — those fixtures are synthetic in-memory packages — but installing it here is fine; `CurrentUser` scope persists to the later generate step.)

One **optional** nit, not a blocker: `powershell-yaml` is installed unpinned while `Pester` is pinned `-MinimumVersion 5.5`. If you want CI fully reproducible, consider a floor here too. No correctness impact.

---

scripts/modules/NeoIPC-Tools/Private/DataDictionary.ps1
```diff
@@ -0,0 +1,674 @@
...
+        if (-not [string]::IsNullOrWhiteSpace($enr)) { return $enr }
+    }
+    return [string]$Stage['executionDateLabel']
+}
+
+function Get-NeoIPCEventDateCode {
+    # Deterministic synthetic code for a stage's event-date field (stages carry no code column of their own).
+    [OutputType([string])]
+    param([Parameter(Mandatory)][System.Collections.IDictionary]$Stage)
+    $slug = ([string]$Stage['name']).ToUpperInvariant() -replace '[^A-Z0-9]+', '_'
```
Brar, line 215: **Latent (non-blocking): synthesized event-date code has no uniqueness guard.**

The event-date `Code` is derived by slugifying the stage name (uppercase → non-alphanumerics collapsed to `_`). Two stages whose names differ only in punctuation/spacing would mint the **same** `NEOIPC_EVENTDATE_<slug>`, producing two Variables rows with an identical, supposedly-unique `Code`.

Can't happen with the current NEOIPC_CORE stages, and the test asserts uniqueness for the fixture — but nothing enforces it for future stage names, and `Code` is presented as the dictionary's stable identifier. If you want to harden it cheaply, either derive the code from a stable per-stage field if one ever exists, or assert no collision across the generated event-date codes (fail-loud, the way the rest of the pipeline does) rather than silently emitting a duplicate.

---

scripts/modules/NeoIPC-Tools/Private/DataDictionary.ps1
```diff
@@ -0,0 +1,674 @@
...
+        foreach ($r in (Get-NeoIPCPatientAttributeRow -Program $program -Index $index)) { $variableRows.Add($r) }
+        $stages = Get-NeoIPCDataDictionaryProgramStage -Program $program -Index $index
+        foreach ($stage in $stages) {
+            foreach ($r in (Get-NeoIPCEventVariableRow -Stage $stage -Program $program -Index $index)) { $variableRows.Add($r) }
+        }
+        foreach ($r in (Get-NeoIPCFormsAndDatesRow -Stage $stages -Program $program)) { $formRows.Add($r) }
+    }
+    $codeListRows = Get-NeoIPCCodeListRow -Package $Package -Index $index
+    $codeListCount = @($codeListRows | ForEach-Object { $_['Code list code'] } | Select-Object -Unique).Count
+    $aboutRows = Get-NeoIPCDataDictionaryAboutRow -Program $programs[0] `
```
Brar: **Latent (non-blocking): the About sheet documents only the first program.**

The orchestrator loops over **all** `$programs` for the Variables and Forms sheets, but the About cover takes its `Program` identity / version from `$programs[0]` only. If a documented package ever carried more than one program, the Variables/Forms/Code-list rows would span every program while the About cover named just the first.

It can't occur today — the assembled package is single-program (the closure is seeded at `NEOIPC_CORE`) — so this is purely latent. Flagging it because the surrounding code is written to handle multiple programs, which makes the single-program About read as an oversight rather than an intentional 1-program assumption. Either assert `programs.Count -eq 1` up front, or list every program on the About sheet.

---

Brar (commented):
## Code review — data-dictionary generator + CI fixes

Full max-effort review of PR #77, covering both the original generator (`93ee25c`) and the four CI-fix commits (`4f93c0d`, `7a5853c`, `07dafda`, `f6087a5`). I ran four finder angles + a sweep over the 674-line `Private/DataDictionary.ps1`, the public cmdlet, wrapper, tests, csproj, psd1, and the CI workflow, then verified every candidate against the code, the test suite, and live upstream (the GitHub runner image, RubyGems, the Ubuntu archive).

### CI-fix commits — all correct ✅
Verified individually (details inline on `build.yml`): the `libgdk-pixbuf-2.0-dev` rename is the right package for 26.04; `mathematical:1.7.0` exists; `sudo env CMAKE_POLICY_VERSION_MINIMUM=3.5` is the correct way to push the CMake-4 policy override through `sudo`'s env-reset; and `powershell-yaml` is genuinely needed by the generate step (the assembler reads YAML). One optional nit: `powershell-yaml` is installed unpinned while Pester is pinned.

### Generator — no confirmed correctness bug
Two **latent, non-blocking** items (both inline), neither triggerable by the current single-program NEOIPC_CORE metadata:

| # | Location | Item |
|---|----------|------|
| 1 | `Private/DataDictionary.ps1:215` | Synthesized event-date `Code` slugifies the stage name with no collision guard — two stage names differing only in punctuation would mint the same code. |
| 2 | `Private/DataDictionary.ps1:434` | About sheet documents only `$programs[0]`, while Variables/Forms span all programs — latent inconsistency if a package is ever multi-program. |

### Verified-and-refuted (recorded so they don't get re-raised)
- **Runner / runtime:** `ubuntu-26.04` exists and ships **PowerShell 7.6.2 + .NET 10**, so `#requires -Version 7.6` is satisfied; `artifacts/` is gitignored and CI uploads it.
- **XLSX frozen pane:** `VerticalSplit` is `ySplit` (rows), so `VerticalSplit=1` + `TopLeftCell=A2` correctly freezes the header row — and the workbook is gated by `OpenXmlValidator` (zero errors) in the test suite.
- **Empty option set / dangling refs:** the empty-set path is explicitly tested as intended; the ref-skips are defensive guards on a package the pipeline already validates for closure.
- **Determinism:** sort keys are effectively unique and a test runs the whole pipeline under `tr-TR` vs `en-US` asserting identical order.

### Conventions / quality
Clean: the exported cmdlet has comment-based help, `ConvertTo-NeoIPCCsvField` and the assembler `-PassThru` are reused (not re-implemented), no CLAUDE.md sync issue (neither instruction file was touched), and the Pester suite is genuinely thorough (validator-gated xlsx, culture determinism, CSV byte-shape, inline-option-limit boundary).

**Verdict:** the CI fixes are sound and the generator is solid — the two flagged items are latent hardening, not blockers.

---

.github/workflows/build.yml
Brar: Pinned a `powershell-yaml` floor (`-MinimumVersion 0.4.2`) to match the Pester pin. (534350a)

scripts/modules/NeoIPC-Tools/Private/DataDictionary.ps1
Brar: Added a fail-loud guard: the orchestrator now asserts every Variables `Code` is unique and throws on a collision, so a slugified event-date code can no longer silently duplicate another variable code. Covered by a new test. (534350a)

scripts/modules/NeoIPC-Tools/Private/DataDictionary.ps1
Brar: Made the single-program contract explicit: the orchestrator now asserts `programs.Count -eq 1` (fail-loud) and operates on that one program throughout, removing the multi-program loop that made the single-program About read as an oversight. Tested. (534350a)

Brar:
Addressed the two latent hardening items and the optional CI nit in `534350a` — CI is green and the new guards are covered by tests:

- **Single-program** (`DataDictionary.ps1`): now a fail-loud assertion (`programs.Count -eq 1`), and the orchestrator operates on that one program throughout — the misleading multi-program loop is gone, so the About cover and the row sheets describe the same program by construction.
- **Event-date code uniqueness** (`DataDictionary.ps1`): the orchestrator now throws on any duplicate Variables `Code`, so a slugified event-date code can never silently collide with another variable's code.
- **`powershell-yaml`** (`build.yml`): pinned a version floor to match the Pester pin.

Also bumped the workflow actions to their Node-24 majors (`checkout@v7` / `setup-dotnet@v5` / `upload-artifact@v7`), clearing the deprecation annotation.

All three inline threads resolved.

Brar:
Re-reviewed `534350a` — all four items verified, nothing further from me:

- **Single-program** — fail-loud `programs.Count -gt 1` guard, the multi-program loop removed, and the cover sheet + row sheets now operate on the one resolved `$program` by construction. ✔
- **Event-date code uniqueness** — the Ordinal `HashSet` over the assembled `Code` column throws on any collision, so a slugified event-date code can't silently duplicate another variable's code. Fail-loud, matching the rest of the pipeline. ✔
- **`powershell-yaml`** — floored at 0.4.2, consistent with the Pester pin. ✔
- **Action bumps** — independently confirmed `actions/checkout@v7`, `actions/setup-dotnet@v5`, and `actions/upload-artifact@v7` all resolve, so the green CI isn't riding a bad tag. ✔

Both new guards are covered by tests, including the nice `'Visit A'` vs `'Visit/A'` slug-collision case. LGTM.

---

PR #78, opened 2026-06-26
Wire metadata import and restructure distribution packaging

Brar:
Completes the metadata pipeline's import half and reworks how the importable packages are distributed.

**Import wiring** — a private DHIS2 POST client and the public `Import-NeoIPCMetadata` cmdlet: dry-run via `importMode=VALIDATE`, returning a normalized import summary.

**Distribution packaging** — the importable packages are now rendered on demand rather than committed. `New-NeoIPCMetadataDistribution.ps1` takes the version explicitly (no default; the new `VERSION` file is the single source of truth); the packages ship as a CI build artifact and as an asset on a manually-published, alpha-marked Release; the undiffable rendered JSON is git-ignored; and CI validates the release-tag format before building.

**Release guards** — on a published Release, CI verifies the tag matches the committed `VERSION` (no tag/file drift) and then auto-advances `VERSION` to the next patch (keeping any pre-release suffix), so the working version always leads the last release.

**Fix** — single-member antibiotic option groups serialized their `options` as a scalar object the DHIS2 importer rejects (surfaced by a live dry-run). They now serialize as a one-element array, with a regression test.

**Docs** — the metadata pipeline's design rationale (locked decisions, diffability principles, capability matrix, verification gates) is distilled into a self-contained `docs/metadata-pipeline-design.md`, linked from the module README.

Validated against the local DHIS2 stack (dry-run HTTP 200, 5808 objects); `Metadata.Tests.ps1` 457/457 green.


.github/workflows/build.yml
```diff
@@ -68,8 +102,69 @@ jobs:
...
+            - name: Advance VERSION to the next patch
+              shell: bash
+              run: |
+                released="${{ github.event.release.tag_name }}"
+                released="${released#v}"
+                if [[ ! "$released" =~ ^([0-9]+)\.([0-9]+)\.([0-9]+)(-.+)?$ ]]; then
+                  echo "::error::Release '$released' is not MAJOR.MINOR.PATCH[-pre]; cannot compute the next version."
+                  exit 1
+                fi
+                next="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}.$((BASH_REMATCH[3] + 1))${BASH_REMATCH[4]}"
```
Brar, line 153: **Highest-impact: the auto-bump can silently downgrade `VERSION`.**

`next` is computed from the released tag and written to main's `VERSION`, but the only guard (line 161) skips when `current == next`. It does **not** guard against main already being *ahead* of `next`.

Sequence: main's `VERSION` is hand-bumped to `0.2.0-alpha` (minor in progress); a late patch release `v0.1.1-alpha` is published from an older commit → `next = 0.1.2-alpha`, which `!= 0.2.0-alpha`, so the push **overwrites `0.2.0-alpha` with `0.1.2-alpha`** — a silent regression of the single-source-of-truth version, with no error.

Make the write conditional on `next` being strictly greater than the current `VERSION` (a `sort -V` comparison), so a release behind main can never roll it back.

---

.github/workflows/build.yml
```diff
@@ -68,8 +102,69 @@ jobs:
...
+                if [[ "$(tr -d '[:space:]' < ./VERSION)" == "$next" ]]; then
+                  echo "VERSION is already $next; nothing to do."
+                  exit 0
+                fi
+                printf '%s\n' "$next" > VERSION
+                git config user.name "github-actions[bot]"
+                git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+                git add VERSION
+                git commit -m "Bump VERSION to $next after release $released"
+                git push origin main
```
Brar: **`git push origin main` has no rebase/retry — a non-fast-forward leaves `VERSION` un-bumped.**

The job checks out `ref: main`, commits, and pushes with no `fetch`/`rebase`/retry. If anything lands on main between checkout and push (an unrelated commit, or two releases published close together), the push is rejected non-fast-forward, the job goes red, and `VERSION` is never advanced — needing a manual fixup.

A small fetch→rebase→retry loop (or opening a PR for the bump) makes this robust against the race. Pairs with the monotonicity issue above — both want the bump to reconcile with main's current state rather than blindly overwrite/replace.

---

.github/workflows/build.yml
```diff
@@ -8,17 +8,51 @@ on:
...
 jobs:
     build-docs:
         # Pinned to an explicit image for reproducibility; 26.04 provides the .NET 10 / PowerShell 7.6
         # runtime the NeoIPC-Tools module and data-dictionary generator require.
         runs-on: ubuntu-26.04
+        # contents: write lets a manually-published Release receive the rendered metadata package as an asset.
+        permissions:
+          contents: write
         steps:
+            - name: Validate version tag format
```
Brar, line 25: **Behavior change: any non-version tag push now hard-fails the build.**

This step runs on every tag push (`startsWith(github.ref, 'refs/tags/')` under the existing `push: tags: ['*']` trigger), so a tag that doesn't match `v\d+.\d+.\d+[-(alpha|beta|rc)[.N]]` now fails the whole job at the gate. Previously `tags: ['*']` built docs for *any* tag.

If only version tags are ever pushed this is harmless — but it's a narrowing worth confirming is intentional (a `nightly`/`deploy-*` marker tag or a typo'd tag would now break CI instead of building). If non-version tags are a possibility, scope the gate to release events only, or to a tag glob like `v*`.

---

scripts/modules/NeoIPC-Tools/Private/DHIS2Http.ps1
```diff
@@ -134,3 +134,87 @@ function Invoke-NeoIPCDhis2Delete {
...
+        [string]$Hostname = 'neoipc.charite.de',
+        [Nullable[int]]$Port = $null,
+
+        [string]$Body,
+        [string]$ContentType = 'application/json',
+        [hashtable]$QueryParameters
+    )
+
+    $effectivePort = if ($null -ne $Port) { $Port } else { -1 }
+    $uriBuilder = [UriBuilder]::new($Scheme, $Hostname, $effectivePort, $Path)
```
Brar: **Reuse: this is the third near-verbatim copy of the URI-build + query-encode + auth block.**

`Invoke-NeoIPCDhis2Post` repeats the `UriBuilder` construction, the `QueryParameters` encode loop, and the Token/Basic credential block already present verbatim in `Invoke-NeoIPCDhis2Get` (lines ~24/56) and `Invoke-NeoIPCDhis2Delete` (lines ~96/104).

Maintenance hazard: a fix to URI/query escaping or to auth (a header change, a credential tweak) now has to be made in three places and can drift between the verbs. Extract a shared `New-NeoIPCDhis2Uri` (scheme/host/port/path + query) and a `Set-NeoIPCDhis2Auth` (mutates the Invoke-RestMethod splat) that all three call. Not blocking, but the third copy is the right moment to factor it.

---

.github/workflows/build.yml
```diff
@@ -68,8 +102,69 @@ jobs:
                 # metadata/ directory alone — no DHIS2 API.
                 ./scripts/New-NeoIPCDataDictionary.ps1 -OutputDirectory ./artifacts/data-dictionary
 
+            - name: Render metadata distribution packages
```
Brar, line 108: **Efficiency: the full metadata package is assembled 2–3× per CI run.**

The data-dictionary step (line 103, `Export-NeoIPCDataDictionary`) assembles the package once; this render step assembles it again — and `New-NeoIPCMetadataDistribution` renders *both* the install-base and the play variant, each its own fresh assembly. So `New-NeoIPCMetadataPackage` runs ~3× per build, each time re-parsing `metadata/`, recomputing the dependency closure, and re-running the ontology/antibiotic generators (3500+ options).

That's several minutes of avoidable CI wall-clock. Both consumers accept a `-Package`, so assemble once up front and pass it in (or at least share one assembly between the base and play renders). Cost-only, not a correctness issue.

---

scripts/modules/NeoIPC-Tools/Public/Metadata.ps1
```diff
@@ -578,6 +578,107 @@ function New-NeoIPCMetadataPackage {
...
+    .PARAMETER Auth
+        Auth hashtable from Resolve-NeoIPCAuth (Token or Basic).
+    .PARAMETER ImportStrategy
+        DHIS2 importStrategy: CREATE_AND_UPDATE (default), CREATE, UPDATE, or DELETE.
+    .PARAMETER AtomicMode
+        DHIS2 atomicMode: ALL (default — all-or-nothing) or NONE (import what is valid, report the rest).
+    .PARAMETER DryRun
+        Validate only (importMode=VALIDATE); the server commits nothing.
+    #>
+    [CmdletBinding(SupportsShouldProcess, DefaultParameterSetName = 'Path')]
```
Brar: **Altitude/safety: a COMMIT import to a live DHIS2 prompts nothing by default.**

`Import-NeoIPCMetadata` declares `SupportsShouldProcess` but no `ConfirmImpact`, so it defaults to `Medium` — below the default `$ConfirmPreference` of `High` — meaning a real (non-`-DryRun`) `CREATE_AND_UPDATE`/`DELETE` import runs without any confirmation prompt unless the user explicitly passes `-Confirm`.

The cmdlet's own help says it "must not be pointed at the production or deployed-test API", yet a mistyped `-Hostname`/`-Auth` on a COMMIT would write to whatever resolves, silently. For a write that can create/update/delete metadata on a live instance, `ConfirmImpact = 'High'` (prompt by default, suppressible with `-Confirm:$false`/`-Force`) is the safer altitude. `-DryRun` could keep `Medium`.

---

.github/workflows/build.yml
```diff
@@ -8,17 +8,51 @@ on:
...
+  release:
+    types: [published]
 
 jobs:
     build-docs:
         # Pinned to an explicit image for reproducibility; 26.04 provides the .NET 10 / PowerShell 7.6
         # runtime the NeoIPC-Tools module and data-dictionary generator require.
         runs-on: ubuntu-26.04
+        # contents: write lets a manually-published Release receive the rendered metadata package as an asset.
+        permissions:
```
Brar: **Least-privilege: `contents: write` is granted job-wide, on every event.**

This makes the whole `build-docs` job (checkout, po4a, dotnet, Pester, data-dictionary, render) run with a write-scoped `GITHUB_TOKEN` on push, PR, dispatch, and release alike — but only the release-gated `gh release upload` step actually needs write (and `bump-version` already declares its own `contents: write`).

Widening the token to write on every run enlarges the blast radius if any third-party action in the job is compromised. Consider dropping `build-docs` to `contents: read` and moving the upload into a release-only job (or a step-scoped token), so write is held only where and when it's used.

---

.github/workflows/build.yml
```diff
@@ -8,17 +8,51 @@ on:
...
                 submodules: true
 
+            - name: Verify the release tag matches the VERSION file
+              # On a published Release the tag (minus the leading v) must equal the committed VERSION, so a
+              # release is always cut from a commit whose VERSION declares exactly that version — no tag/file drift.
+              if: github.event_name == 'release'
+              shell: bash
+              run: |
+                tag="${{ github.event.release.tag_name }}"
+                file="$(tr -d '[:space:]' < ./VERSION)"
```
Brar: **Latent: `tr -d '[:space:]'` won't strip a UTF-8 BOM from `VERSION`.**

`VERSION` is BOM-free today (verified — it starts `30 2e 31…` = `0.1.0-alpha`), so this is latent only. But `tr -d '[:space:]'` removes whitespace, not the BOM bytes (`EF BB BF`), so if `VERSION` is ever re-saved with a BOM, `file` becomes `<BOM>0.1.0-alpha` and `[[ "${tag#v}" != "$file" ]]` is true even though the visible text matches — blocking the release with a message that reads like a content mismatch. The same applies to the `bump-version` equality check at line 161. Strip the BOM too (e.g. `sed '1s/^\xEF\xBB\xBF//'` or read it BOM-aware) for robustness.

---

Brar (commented):
## Code review — metadata import wiring + distribution packaging

Reviewed across four finder angles + verification, covering the import POST client, `Import-NeoIPCMetadata`, the antibiotic single-member fix, the distribution version change, and the CI release/version automation. **The shipped PowerShell is solid and well-tested** — every candidate raised against the import path was refuted on inspection (see below). The findings that survived are in the CI/release automation (robustness) plus reuse/altitude/least-privilege; none is a crash or data-corruption in the module code path.

### Findings (8 inline)
| # | Location | Kind | Severity |
|---|----------|------|----------|
| 1 | `build.yml:160` (bump-version) | **VERSION auto-bump can silently downgrade** — guards equality only, not "main already ahead" | Med (corrupts source-of-truth version) |
| 2 | `build.yml:170` | `git push origin main` with no rebase/retry — non-fast-forward leaves VERSION un-bumped | Med-low |
| 3 | `build.yml:23` | Any non-version tag push now hard-fails the build (was: build any tag) | Low (behavior change) |
| 4 | `DHIS2Http.ps1:181` | URI/query/auth block now copy-pasted a third time (Get/Delete/Post) | Med (maintainability) |
| 5 | `build.yml:105` | Full package assembled 2–3× per CI run (data-dictionary + base + play) | Med-low (CI cost) |
| 6 | `Metadata.ps1:613` | COMMIT import to live DHIS2 has no default confirm prompt (no `ConfirmImpact=High`) | Low-med (safety) |
| 7 | `build.yml:20` | `contents: write` job-wide on every event; only the release upload needs it | Low (least-privilege) |
| 8 | `build.yml:49` | `tr -d '[:space:]'` won't strip a BOM in the tag-vs-VERSION check (VERSION is BOM-free today) | Latent-low |

The most consequential is **#1** — the version auto-bump rewriting the canonical `VERSION` downward is exactly the kind of single-source-of-truth corruption worth guarding before this runs for real; **#1/#2** want the bump to reconcile with main's current state rather than blindly overwrite.

### Verified-and-refuted (recorded so they don't recur)
- `-WhatIf` "$response null-deref": can't happen — the cmdlet `return`s before POSTing (test-proven, 0 invocations).
- Array/`$null` body in the response normalizer: `.PSObject.Properties` is safe, yields graceful nulls, no crash.
- BOM in the POST payload: `[System.IO.File]::ReadAllText` strips the BOM, so none reaches the server.
- Status precedence (`.response.status` first): correct, proven by the 409-ERROR fixture (surfaces ERROR over the top-level WARNING).
- The antibiotic `@()`-at-call-site fix and `UriBuilder.Query` `'?'` handling are both correct.

### Conventions
Clean: `Import-NeoIPCMetadata` is exported with full comment-based help, `Invoke-NeoIPCDhis2Post` stays private, no `Co-Authored-By`, no CLAUDE.md/copilot-instructions drift, and the deleted `metadata/dist/*.json` packages have no dangling consumers (README/docs use `<version>` placeholders). One borderline: the `bump-version` job's `git push origin main` technically matches the repo's "never push directly to main" rule, but the actor is `github-actions[bot]` in maintainer-authored release automation (the comment even flags the branch-protection bypass it needs) — that reads as an intentional maintainer decision, not the dev/agent push the rule targets. Worth a conscious confirmation rather than a blocker.

Non-blocking overall — #1 is the one I'd resolve (or consciously accept) before the release automation runs.

---

Brar:
Addressed in e642d5c — full NeoIPC-Tools suite 481/481 green, and the bump / tag / BOM bash logic unit-checked locally.

- **#1 / #2** — the bump now only ever *advances* (a `sort -V` monotonic guard, so a patch released from an older commit or a hand-bumped minor in progress is never rolled back) and re-syncs to `main` + retries on a non-fast-forward push.
- **#3** — the tag-format gate is scoped to `v*` tags, so a non-version (deploy/marker) tag builds as before.
- **#4** — the URI-build + auth blocks are factored into `New-NeoIPCDhis2Uri` / `Set-NeoIPCDhis2Auth`, shared by GET/DELETE/POST.
- **#6** — `Import-NeoIPCMetadata` is `ConfirmImpact = 'High'`, so a real COMMIT prompts by default; a `-DryRun` stays quiet (it only validates).
- **#7** — `build-docs` runs with a read-only token now; the Release-asset upload moved to its own write-scoped, release-gated job (top-level default `contents: read`).
- **#8** — both VERSION reads strip a UTF-8 BOM.

**Deferred — #5** (assemble-once CI cost): real, but `New-NeoIPCMetadataDistribution` renders distinct install-base + play assemblies, so sharing one with the data-dictionary step needs the generator refactored to accept a pre-assembled `-Package`. A focused follow-up — cost, not correctness — so out of this PR.

.github/workflows/build.yml
Brar: Fixed in e642d5c — the bump now guards with `sort -V` and only ever advances, so a patch released from an older commit (or a hand-bumped minor/major in progress) can no longer roll `VERSION` back; it logs and skips instead.

.github/workflows/build.yml
Brar: Fixed in e642d5c — the bump re-syncs to the tip of `main` (fetch + reset --hard) each attempt and retries the push up to 3x on a non-fast-forward, so a commit landing on main between checkout and push no longer drops the bump.

.github/workflows/build.yml
Brar: Fixed in e642d5c — the gate is scoped to `v*` tags (`refs/tags/v`), so a non-version deploy/marker tag builds as before; releases and `v*` tags are still validated.

scripts/modules/NeoIPC-Tools/Private/DHIS2Http.ps1
Brar: Fixed in e642d5c — extracted `New-NeoIPCDhis2Uri` (scheme/host/port/path + query) and `Set-NeoIPCDhis2Auth` (token/basic, opt-in unencrypted); GET/DELETE/POST all route through them now, so a URI or auth change lands once.

.github/workflows/build.yml
Brar: Deferred to a focused follow-up (captured in the project task tracking) — the render assembles distinct install-base and play packages, so sharing one assembly with the data-dictionary step needs `New-NeoIPCMetadataDistribution` refactored to accept a pre-assembled `-Package`. Cost, not correctness, so out of this PR.

scripts/modules/NeoIPC-Tools/Public/Metadata.ps1
Brar: Fixed in e642d5c — `Import-NeoIPCMetadata` now declares `ConfirmImpact` High, so a real (non-dry-run) COMMIT prompts by default; a `-DryRun` stays quiet (it only validates). The seed and tests pass `-Confirm:$false`.

.github/workflows/build.yml
Brar: Fixed in e642d5c — top-level default is now `contents: read`, `build-docs` runs read-only, and the Release-asset upload moved to its own write-scoped, release-gated job (beside the existing version-bump job).

.github/workflows/build.yml
Brar: Fixed in e642d5c — both `VERSION` reads strip a leading UTF-8 BOM before comparing (sed BOM-strip + tr -d whitespace).

Brar:
Re-reviewed `e642d5c` — all seven hold up, and the `#5` deferral is the right call:

- **#1 / #2** — the bump loop re-syncs (`fetch` + `reset --hard origin/main`), writes only when `next` is the strict `sort -V` maximum (a hand-bumped minor or an older-commit patch release is never rolled back), and retries 3× on a non-fast-forward → errors out rather than silently dropping the bump. ✔
- **#3** — `startsWith(github.ref, 'refs/tags/v')`, so a non-version marker/deploy tag builds as before. ✔
- **#4** — `New-NeoIPCDhis2Uri` / `Set-NeoIPCDhis2Auth` shared by GET/DELETE/POST; behavior preserved (POST keeps `AllowUnencrypted`, GET/DELETE don't, as before). ✔
- **#6** — `ConfirmImpact='High'` + the `-DryRun` → `$ConfirmPreference='None'` suppression; the COMMIT-mode tests now pass `-Confirm:$false`, dry-run/WhatIf correctly don't. ✔
- **#7** — the part I checked hardest, since splitting the upload into its own job risks dropping the artifact handoff: it's correct — `release-assets` pulls build-docs's output via `download-artifact` (the `path: ./artifacts/` → `./artifacts/metadata-package/*.json` round-trip lines up) and sets `GH_REPO` so the checkout-less `gh release upload` resolves the repo. build-docs is now read-only under the top-level `contents: read` default; only the two release-gated jobs request write. ✔
- **#8** — BOM stripped on both VERSION reads. ✔

**#5** (assemble-once) — agreed it's a separate refactor (teach the dist generator to accept a pre-assembled `-Package`), cost not correctness, fine to defer. LGTM.

---

PR #80, opened 2026-06-30
Report logging via the shared logger, plus metadata-import round-trip verification

Brar:
Two related strands of work on the reports pipeline and the metadata tooling.

### Report logging
Routes the reports through the shared `logger`-based logging (new `reports/common/logging.R`), so report runs emit structured, level-controlled diagnostics consistent with neoipcr's logging.

- Unifies log-verbosity control across all reports and retires the Partner-Report-specific debug parameter.
- Resolves log-message placeholders in the caller's frame.

### Metadata-import verification
Strengthens the metadata round-trip verifier in the `NeoIPC-Tools` module (`MetadataVerify.ps1`, with Pester coverage):

- Connects same-payload owned references on import and verifies them.
- Surfaces the DHIS2 import error message when the server returns a bare WebMessage, instead of a generic failure.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,7 +67,7 @@ Reports live under `reports/`:
 
 ### Shared Infrastructure
 
-- **Shared R code**: `reports/common/` — `helpers.R` (locale parsing, string resource loading, DHIS2 connection helpers), `load-neoipcr.R`, `parse-args.R` (CLI arg parsing), `getDataset.R` (dataset export), `reference.docx` (Word template)
+- **Shared R code**: `reports/common/` — `helpers.R` (locale parsing, string resource loading, DHIS2 connection helpers), `load-neoipcr.R`, `parse-args.R` (CLI arg parsing), `getDataset.R` (dataset export), `logging.R` (unified `logger`-based logging: `configure_logging()` + `logInfo`/`logVerbose`/`logDebug`/`logWarn`/`logError`), `reference.docx` (Word template)
 - **Base string resources**: `reports/common.yaml` (English domain terms, table headers, footnotes)
 - **Pandoc filters**: `reports/filters/pandoc-quotes.lua` (language-aware typographic quotes), `remove-empty-sections.lua`
 
@@ -264,6 +264,18 @@ No `sprintf` `%s`, markdown, or LaTeX syntax in translatable strings. Use `glue`
 ### PowerShell Scripts
 
 Approved PS verbs + PascalCase noun (e.g., `New-PartnerReports.ps1`). All wrappers in `scripts/`. Shared helpers in `scripts/NeoIPCReportHelpers.ps1` (dot-sourced).
+
+### Logging
+
+All report R code and neoipcr log through the `logger` package (`reports/common/logging.R`). Three R namespaces —
+the report's slug (e.g. `partner-report`), `report-common` (the shared `common/` layer), and `neoipcr` — let every
+line self-identify its source. Verbosity is **one** setting (`quiet`/`normal`/`verbose`/`debug`), propagated to the
+R / Quarto child processes via the **`NEOIPC_LOG_LEVEL`** environment variable: the default `normal` shows lifecycle
+progress; `verbose`/`debug` reveal the DHIS2 query trace (URL + HTTP status + row count — **never** response bodies,
+a data-protection boundary). The `New-*.ps1` wrappers map the standard `-Quiet`/`-Verbose`/`-Debug` switches to it
+(via `Invoke-WithNeoIPCAuth -ExtraEnvVars`); the `Generate-*Data.R` wrappers' `--quiet`/`--verbose`/`--debug` flags
+and Partner-Report's `-P debug:true` override it. When `NEOIPC_LOG_FILE` is set (by the NeoIPC-Reporting .NET
+service), the R side writes structured JSON to that file instead of the console.
 
 ### Argument Handling
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -273,8 +273,8 @@ line self-identify its source. Verbosity is **one** setting (`quiet`/`normal`/`v
 `verbose`/`debug` reveal the DHIS2 query trace (URL + HTTP status + row count — **never** response bodies, a
 data-protection boundary). The `New-*.ps1` wrappers map the standard `-Quiet`/`-Verbose`/`-Debug` switches to it and
 pass it to the children **two** ways: the **`NEOIPC_LOG_LEVEL`** environment variable (read by the QMDs and neoipcr)
-and the native `--quiet`/`--verbose`/`--debug` CLI flags on the `Generate-*Data.R` / `quarto render` calls; `-Quiet`
-additionally silences the wrapper's own progress/verbose streams. Each `Generate-*Data.R` resolves a native CLI flag
+and native CLI flags — `--quiet`/`--verbose`/`--debug` on the `Generate-*Data.R` calls and `--quiet`/`--log-level`
+on `quarto render`; `-Quiet` additionally silences the wrapper's own progress/verbose streams. Each `Generate-*Data.R` resolves a native CLI flag
 first, falls back to `NEOIPC_LOG_LEVEL` (so the .NET service can drive it environment-only), and republishes the
 resolved level for neoipcr and any child processes. When `NEOIPC_LOG_FILE` is set (by the NeoIPC-Reporting .NET
 service), the R side writes structured JSON to that file instead of the console.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -269,12 +269,14 @@ Approved PS verbs + PascalCase noun (e.g., `New-PartnerReports.ps1`). All wrappe
 
 All report R code and neoipcr log through the `logger` package (`reports/common/logging.R`). Three R namespaces —
 the report's slug (e.g. `partner-report`), `report-common` (the shared `common/` layer), and `neoipcr` — let every
-line self-identify its source. Verbosity is **one** setting (`quiet`/`normal`/`verbose`/`debug`), propagated to the
-R / Quarto child processes via the **`NEOIPC_LOG_LEVEL`** environment variable: the default `normal` shows lifecycle
-progress; `verbose`/`debug` reveal the DHIS2 query trace (URL + HTTP status + row count — **never** response bodies,
-a data-protection boundary). The `New-*.ps1` wrappers map the standard `-Quiet`/`-Verbose`/`-Debug` switches to it
-(via `Invoke-WithNeoIPCAuth -ExtraEnvVars`); the `Generate-*Data.R` wrappers' `--quiet`/`--verbose`/`--debug` flags
-and Partner-Report's `-P debug:true` override it. When `NEOIPC_LOG_FILE` is set (by the NeoIPC-Reporting .NET
+line self-identify its source. Verbosity is **one** setting (`quiet`/`normal`/`verbose`/`debug`): the default `normal` shows lifecycle progress;
+`verbose`/`debug` reveal the DHIS2 query trace (URL + HTTP status + row count — **never** response bodies, a
+data-protection boundary). The `New-*.ps1` wrappers map the standard `-Quiet`/`-Verbose`/`-Debug` switches to it and
+pass it to the children **two** ways: the **`NEOIPC_LOG_LEVEL`** environment variable (read by the QMDs and neoipcr)
+and the native `--quiet`/`--verbose`/`--debug` CLI flags on the `Generate-*Data.R` / `quarto render` calls; `-Quiet`
+additionally silences the wrapper's own progress/verbose streams. Each `Generate-*Data.R` resolves a native CLI flag
+first, falls back to `NEOIPC_LOG_LEVEL` (so the .NET service can drive it environment-only), and republishes the
+resolved level for neoipcr and any child processes. When `NEOIPC_LOG_FILE` is set (by the NeoIPC-Reporting .NET
 service), the R side writes structured JSON to that file instead of the console.
 
 ### Argument Handling
--- a/reports/Partner-Certificate/_setup.qmd
+++ b/reports/Partner-Certificate/_setup.qmd
@@ -1,8 +1,14 @@
 ```{r}
 #| label: Setup
+#| warning: false
 source("../common/helpers.R")
 source("../common/logging.R")
 configure_logging(report = "partner-certificate")
+# R warnings (e.g. from the import_dhis2 fallback below) are captured into the
+# unified log channel by configure_logging()'s warning hook; the chunk's
+# `warning: false` keeps them out of the rendered certificate body. (The import
+# runs inside this chunk, so a chunk option is needed rather than
+# opts_chunk$set, which only affects subsequent chunks.)
 
 locale <- Sys.getenv("LC_ALL")
 localeObj <- parse_locales(locale)[[1]]
--- a/reports/Partner-Certificate/_setup.qmd
+++ b/reports/Partner-Certificate/_setup.qmd
@@ -1,6 +1,8 @@
 ```{r}
 #| label: Setup
 source("../common/helpers.R")
+source("../common/logging.R")
+configure_logging(report = "partner-certificate")
 
 locale <- Sys.getenv("LC_ALL")
 localeObj <- parse_locales(locale)[[1]]
@@ -50,7 +52,7 @@ if (is.null(data$startYear) || is.null(data$endYear) || is.null(data$hospitalNam
     port = params$dhis2Port, path = params$dhis2Path)
 
   # Import date from our DHIS2 server
-  ds <- suppressWarnings(neoipcr::import_dhis2(
+  ds <- neoipcr::import_dhis2(
     connection_options = connection_options,
     dataset_options = neoipcr::dhis2_dataset_options(
       surveillance_end_from = startFrom,
@@ -62,7 +64,7 @@ if (is.null(data$startYear) || is.null(data$endYear) || is.null(data$hospitalNam
                            "delivery_mode", "siblings"),
       include_enrollment = "full",
       include_event = "full"
-      )))
+      ))
 
   department <- ds$metadata$departments |>
     dplyr::filter(code == data$departmentCode) |>
--- a/reports/Patient-Data-Report/_setup.qmd
+++ b/reports/Patient-Data-Report/_setup.qmd
@@ -7,6 +7,12 @@ source("../common/logging.R")
 suppressMessages(library(tidyverse, warn.conflicts = FALSE, quietly = TRUE))
 configure_logging(report = "patient-data-report")
 
+# R warnings (e.g. from import_dhis2) are captured into the unified log channel
+# by configure_logging()'s warning hook; keep them out of the rendered report
+# body. Set here in the setup chunk so it applies to the later import/filter
+# chunks (mirrors Partner-Report).
+knitr::opts_chunk$set(warning = FALSE)
+
 locale <- Sys.getenv("LC_ALL")
 invisible(Sys.setlocale(category = "LC_ALL", locale = locale))
 localeObj <- parse_locales(locale)[[1]]
--- a/reports/Patient-Data-Report/_setup.qmd
+++ b/reports/Patient-Data-Report/_setup.qmd
@@ -3,7 +3,9 @@
 source("../common/load-neoipcr.R")
 suppressMessages(load_neoipcr(dev_pkg_path = "../../../neoipcr"))
 source("../common/helpers.R")
+source("../common/logging.R")
 suppressMessages(library(tidyverse, warn.conflicts = FALSE, quietly = TRUE))
+configure_logging(report = "patient-data-report")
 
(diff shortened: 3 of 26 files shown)


reports/Patient-Data-Report/_setup.qmd
```diff
@@ -21,7 +23,7 @@ connection_options <- get_connection_options(
   scheme = params$dhis2Scheme, hostname = params$dhis2Hostname,
   port = params$dhis2Port, path = params$dhis2Path)
 
-ds <- suppressWarnings(neoipcr::import_dhis2(
+ds <- neoipcr::import_dhis2(
```
Brar, line 32: **IMPORTANT (warnings can now leak into the rendered body):** Removing `suppressWarnings()` here delegates warning capture to `configure_logging()`'s `logger::log_warnings(muffle = FALSE)` hook — but `muffle = FALSE` means warnings are logged **and still propagate** to knitr. In Partner-Report that's safe because its `_setup.qmd` sets both `#| warning: false` and `knitr::opts_chunk$set(warning = FALSE)` (with a comment: *"R warnings are captured into the unified log channel … keep them out of the rendered report body"*). This report has **neither** — I checked: the "Data Import" chunk has no `#| warning: false`, there's no `opts_chunk$set(warning = FALSE)` anywhere in this setup, and no project-level `warning: false` in `_quarto.yml`. knitr defaults `warning = TRUE`, so any warning `import_dhis2` raises now renders as a `## Warning:` block in the delivered PDF where `suppressWarnings()` previously hid it.

You clearly knew the right pattern (you applied it in Partner-Report) — just mirror it here: add `#| warning: false` to this chunk, or `knitr::opts_chunk$set(warning = FALSE)` in this setup. Same issue in `Partner-Certificate/_setup.qmd` (see other comment).

---

reports/Partner-Certificate/_setup.qmd
```diff
@@ -50,7 +52,7 @@ if (is.null(data$startYear) || is.null(data$endYear) || is.null(data$hospitalNam
     port = params$dhis2Port, path = params$dhis2Path)
 
   # Import date from our DHIS2 server
-  ds <- suppressWarnings(neoipcr::import_dhis2(
+  ds <- neoipcr::import_dhis2(
```
Brar, line 61: **IMPORTANT (same warning-leak as Patient-Data-Report):** `suppressWarnings()` is removed here too, but this "Setup" chunk has no `#| warning: false` and there's no `opts_chunk$set(warning = FALSE)` in this file (nor a project-level setting). With `log_warnings(muffle = FALSE)`, warnings from `import_dhis2` now propagate into the rendered certificate body. Mirror the Partner-Report treatment (`#| warning: false`) so the new "log, don't render" behaviour is uniform across all four reports.

---

scripts/modules/NeoIPC-Tools/Public/Metadata.ps1
```diff
@@ -667,9 +677,36 @@ function Import-NeoIPCMetadata {
     $status = if ($report -and ($report.PSObject.Properties.Name -contains 'status')) { $report.status }
     elseif ($body -and ($body.PSObject.Properties.Name -contains 'status')) { $body.status } else { $null }
 
-    Write-Verbose ("metadata import ({0}): HTTP {1}, status {2}{3}." -f `
+    # When the import fails at the persistence layer (e.g. a Hibernate not-null/transient flush error),
+    # DHIS2 returns a bare WebMessage with a top-level `message` and NO `.response`/typeReports node — so
+    # the structured-report fields above are all null and the real cause hides in `message`. Surface it so
+    # callers and logs name the actual fault instead of an opaque "status ERROR (HTTP 409)".
+    $errorMessage = if (-not $hasResponse -and $body) {
```
Brar, line 684: **IMPORTANT (new behaviour ships untested):** This bare-WebMessage `ErrorMessage` surfacing (commit 469f818) and the `-ConnectReferences` connect pass below at line 702 (commit 25452a3) are net-new branches in `Import-NeoIPCMetadata`, but `Tests/Metadata.Tests.ps1` has no assertion for `ErrorMessage`, `ConnectPassStatus`, or `-ConnectReferences` (grep returns zero matches). The existing "reports an ERROR import (HTTP 409)" test uses a `.response`-wrapped fixture, which takes the `$hasResponse = $true` path and never exercises this branch — so the headline behaviour of two commits is uncovered, while the sibling `MetadataVerify.ps1` got a thorough 15-test suite.

Suggested coverage: (a) a bare WebMessage `{ httpStatusCode:409, status:"ERROR", message:"…", devMessage:"…" }` fixture asserting `$r.ErrorMessage` is the joined `message` / `devMessage`; (b) a happy-path assertion that `ErrorMessage` is `$null`; (c) a `-ConnectReferences` test asserting `Invoke-NeoIPCDhis2Post` runs twice on OK/WARNING and the connect pass is skipped on a `$null`/ERROR status.

---

scripts/modules/NeoIPC-Tools/Public/MetadataVerify.ps1
```diff
@@ -0,0 +1,332 @@
...
+                    $detail = Get-NeoIPCOrderDriftDetail $expIds $actIds
+                    if ($detail) {
+                        $records.Add([pscustomobject]@{ Type = $Type; Id = $id; Code = $code; Kind = 'OrderDrift'; Field = $fname; Detail = $detail })
+                    }
+                }
+                continue
+            }
+            $e = $expected | ConvertTo-Json -Compress -Depth 20
+            $a = $actual | ConvertTo-Json -Compress -Depth 20
+            if ($e -cne $a) {
```
Brar, line 239: **IMPORTANT (order-sensitive JSON equality can false-positive):** This `FieldMismatch` branch compares nested objects via `ConvertTo-Json -Compress` string equality (`-cne`), which is **property-order-sensitive**. Multi-key nested fields that aren't references and aren't `*Array` types fall here — the type maps carry exactly such fields: `renderType` (`{ DESKTOP, MOBILE, … }`) on `programStageDataElements`/`programTrackedEntityAttributes`/`programStageSections`, and `style` (`{ icon, … }`) on `programStages`. An `[ordered]` package hashtable and a DHIS2 `PSCustomObject` holding identical data in a different key order serialize to different strings and report a spurious `FieldMismatch`. Since the verifier is a clean-round-trip gate, that would block a perfectly good seed.

Failure case: DHIS2 returns `renderType` as `{"MOBILE":…,"DESKTOP":…}` while the package emitted `{"DESKTOP":…,"MOBILE":…}` → `FieldMismatch` on an identical value. Suggested fix: canonicalize before comparing — recursively sort keys, or diff the parsed objects key-by-key — so only genuine value differences surface. (Scalars are unaffected; only multi-key nested objects need this.)

---

reports/common/logging.R
```diff
@@ -0,0 +1,118 @@
...
+    verbose = logger::DEBUG,
+    debug   = logger::TRACE,
+    logger::INFO)
+}
+
+# Configure logging for a report run.
+#
+# `report` is the report slug used as this report's logger namespace (e.g.
+# "partner-report"); NULL (the default) uses "report-common", for standalone
+# common entry points such as getDataset.R. `verbosity` is one of
```
Brar: **NIT (stale doc reference):** This cites `getDataset.R` as the motivating example for the `report = NULL` default, but `getDataset.R` doesn't `source("../common/logging.R")`, never calls `configure_logging()`, and uses `cat()` for all output — so nothing currently uses this NULL path. Per the CLAUDE.md rule to verify factual claims in comments against the source, either drop the `getDataset.R` example (the NULL default reads fine as "standalone common entry points") or wire `getDataset.R` through `configure_logging()` if that was the intent.

---

CLAUDE.md
```diff
@@ -265,6 +265,20 @@ No `sprintf` `%s`, markdown, or LaTeX syntax in translatable strings. Use `glue`
...
 
+### Logging
+
+All report R code and neoipcr log through the `logger` package (`reports/common/logging.R`). Three R namespaces —
+the report's slug (e.g. `partner-report`), `report-common` (the shared `common/` layer), and `neoipcr` — let every
+line self-identify its source. Verbosity is **one** setting (`quiet`/`normal`/`verbose`/`debug`): the default `normal` shows lifecycle progress;
+`verbose`/`debug` reveal the DHIS2 query trace (URL + HTTP status + row count — **never** response bodies, a
+data-protection boundary). The `New-*.ps1` wrappers map the standard `-Quiet`/`-Verbose`/`-Debug` switches to it and
+pass it to the children **two** ways: the **`NEOIPC_LOG_LEVEL`** environment variable (read by the QMDs and neoipcr)
+and the native `--quiet`/`--verbose`/`--debug` CLI flags on the `Generate-*Data.R` / `quarto render` calls; `-Quiet`
```
Brar: **NIT (imprecise wording):** This lumps `quarto render` in with the `--verbose`/`--debug` flag form, but the Quarto children actually receive `--log-level info` / `--log-level debug` (only `--quiet` is shared) — as implemented by `$quartoVerbosityArgs` in all five wrappers. The wrapper inline comments get this right; only this prose is off. Reword to distinguish the two, e.g. *"…native `--quiet`/`--verbose`/`--debug` on `Generate-*Data.R` and `--quiet`/`--log-level` on `quarto render`."* Mirror the fix in `.github/copilot-instructions.md` to keep the two in sync (per the sync guardrail).

---

Brar (commented):
## Review — report logging + metadata-import round-trip verification

Well-engineered PR across two independent strands; build-docs CI is green. No blockers — findings are IMPORTANT-and-below.

**Strand 1 — report logging (R/Quarto/PowerShell):** Clean and correct. `reports/common/logging.R` wires the three documented namespaces (report slug, `report-common`, `neoipcr`), the `.topenv = parent.frame()` glue-frame handling correctly resolves placeholders in the caller's frame, and `NEOIPC_LOG_LEVEL`/`NEOIPC_LOG_FILE` match the CLAUDE.md spec. The retired Partner-Report `debug` parameter is removed with **zero** dangling references — I traced the QMDs (incl. the hand-edited `Partner-Report.de.qmd`), `Generate-PartnerData.R`, and `New-PartnerReports.ps1` clean — and the five `New-*.ps1` wrappers are consistently aligned per the wrapper-alignment guardrail. The one real concern: removing `suppressWarnings()` from the imports was paired with warning suppression only in Partner-Report (and `include: false` chunks in Reference-Report), leaving **Patient-Data-Report and Partner-Certificate** without it → warnings can now render into those two PDFs (inline, confirmed).

**Strand 2 — metadata verification (PowerShell):** Solid. Correct `$null`-on-LHS comparisons, strict-mode-safe property access, proper param sets, and the bare-WebMessage error parsing won't throw on the happy path. Two concerns: the new `Import-NeoIPCMetadata` behaviours (`ErrorMessage` surfacing + the `-ConnectReferences` connect pass — the headline of two commits) ship with **no test coverage**, in contrast to the thorough 15-test `MetadataVerify` suite; and the verifier's `FieldMismatch` branch compares nested objects (`renderType`, `style`) via order-sensitive `ConvertTo-Json` equality, which can false-positive and block a good seed.

**Praise:**
- The `MetadataVerify.Tests.ps1` suite (15 tests) is non-tautological and encodes real regressions the design depends on — notably "does NOT flag `dataElementGroups.dataElements` reordered" (the `<set>` vs `<list>` distinction) and "round-trips option `sortOrder` by value". Fully mocked from an in-memory state — no live HTTP (verified: 15/15 pass).
- Clean retirement of the Partner `debug` parameter and genuinely consistent wrapper/`Generate-*Data.R` alignment.

**Data-protection boundary:** Confirmed clean — the logging layer never logs DHIS2 response bodies (it only sets the `neoipcr` namespace threshold/appender; `log_dhis2_request()` logs URL + status + row count only), and the verifier's `IgnoreField` default excludes `password`.

Summary of inline items: **(IMPORTANT)** warning leak in 2 reports · missing tests for new import behaviours · order-sensitive nested-object compare in the verifier; **(NIT)** stale `getDataset.R` doc reference · `quarto render` flag-form wording in CLAUDE.md/copilot-instructions. **Mergeable** after the two warning-suppression lines and ideally the import-behaviour tests.

_Posted as a comment review (GitHub doesn't allow approving your own PR). I'll watch for your replies._

---

reports/Patient-Data-Report/_setup.qmd
Brar: Fixed in e82cb97. Added `knitr::opts_chunk$set(warning = FALSE)` to this report's setup chunk (with the same rationale comment as Partner-Report), placed in the setup chunk so it also covers the later **Data Import** and **Filter to patient** chunks. Warnings from `import_dhis2` still reach the unified log channel via the `log_warnings()` hook — they just no longer render into the PDF.

---

reports/Partner-Certificate/_setup.qmd
Brar: Fixed in e82cb97. Used a chunk-level `#| warning: false` on the Setup chunk here rather than `opts_chunk$set` — the `import_dhis2` fallback runs **inside** this same Setup chunk, and `opts_chunk$set` only affects *subsequent* chunks, so the chunk option is what actually suppresses the in-chunk warnings. Added a comment noting that. The "log, don't render" behaviour is now uniform across all four reports.

---

reports/common/logging.R
Brar: Fixed in e82cb97. Dropped the `getDataset.R` example — it doesn't source `logging.R` or call `configure_logging()`, so the reference was stale. The comment now reads "…for standalone entry points in the shared `common/` layer," which keeps the rationale for the `report = NULL` default without naming a file that doesn't use the path.

---

CLAUDE.md
Brar: Fixed in e82cb97. Reworded to distinguish the two flag forms — `--quiet`/`--verbose`/`--debug` on the `Generate-*Data.R` calls and `--quiet`/`--log-level` on `quarto render` — and mirrored the same edit in `.github/copilot-instructions.md` to keep the two in sync per the sync guardrail.

---

Brar:
Thanks for the review. The **Strand-1 (report logging)** items are addressed in e82cb97 (follow-up commit; no force-push, so the threads stay anchored):

- **Warning leak (the one real concern)** — Patient-Data-Report now sets `knitr::opts_chunk$set(warning = FALSE)` in its setup (covering the import + filter chunks), and Partner-Certificate uses a chunk-level `#| warning: false` (its import runs inside the setup chunk, where `opts_chunk$set` wouldn't apply). All four reports now keep R warnings in the unified log channel, out of the rendered body.
- **NIT — stale `getDataset.R` reference** in `logging.R`: dropped (that NULL path is currently unused).
- **NIT — `quarto render` flag wording** in `CLAUDE.md` + `.github/copilot-instructions.md`: corrected so it distinguishes `Generate-*Data.R` (`--quiet`/`--verbose`/`--debug`) from `quarto render` (`--quiet`/`--log-level`).

Both edited QMDs parse cleanly; build-docs CI re-runs on the push.

The two **Strand-2 (metadata-import verification)** items — the untested `Import-NeoIPCMetadata` behaviours and the order-sensitive nested-object compare in the verifier — are being handled separately, so I've left those two threads open.

---

Brar:
Verified the Strand-1 fixes against `e82cb97` — all four resolved cleanly:

- **Warning leak** — Patient-Data-Report sets `knitr::opts_chunk$set(warning = FALSE)` in setup (covers the import + filter chunks); Partner-Certificate uses a chunk-level `#| warning: false`, which is the right choice since its import runs *inside* the Setup chunk where `opts_chunk$set` wouldn't apply. All four reports now keep import warnings in the log channel, out of the rendered body.
- **`getDataset.R` NIT** — stale reference dropped, rationale for the `report = NULL` default preserved.
- **`quarto render` flag wording** — corrected and mirrored identically in `CLAUDE.md` and `.github/copilot-instructions.md`, so the sync guardrail holds.

build-docs is green on the follow-up commit. Sounds good on deferring the two Strand-2 items (untested `Import-NeoIPCMetadata` branches, order-sensitive nested-object compare in the verifier) to separate work — both are real but neither blocks this PR. I'll leave those two threads open as you have them. 👍

---

scripts/modules/NeoIPC-Tools/Public/MetadataVerify.ps1
Brar: Fixed in 19acb94. Both sides are now canonicalized before the compare: a recursive `ConvertTo-NeoIPCCanonical` sorts the keys of every nested object — the package's `[ordered]`/`IDictionary` side and DHIS2's `PSCustomObject` side alike — and preserves array order, so identical data in a different key order (`renderType { DESKTOP, MOBILE }`, `style { icon, … }`) serializes to the same string instead of a spurious `FieldMismatch`. Only object keys are reordered; scalars and array order pass through untouched, so a genuine value difference still surfaces.

Two tests added in `MetadataVerify.Tests.ps1`: a key-reordered `renderType` (`MOBILE`/`DESKTOP` swapped) with identical data is **not** flagged, and a `renderType` whose nested value genuinely differs **is** still flagged — so the canonicalization fixes the false positive without masking real drift. The independent review also confirmed the only old-vs-new outcome change is the benign `package=@()` vs `DHIS2=null` case (an empty list and "no value" being treated as equivalent), which removes noise rather than hiding a drop.

---

scripts/modules/NeoIPC-Tools/Public/Metadata.ps1
Brar: Fixed in 19acb94. Added the suggested coverage to the `Import-NeoIPCMetadata` suite:

- **(a)** a bare WebMessage fixture (`httpStatusCode 409`, `status ERROR`, top-level `message` + `devMessage`, **no** `.response`) asserts `ErrorMessage` is the joined `message / devMessage`;
- **(b)** a happy-path (WebMessage-wrapped) import asserts `ErrorMessage` is null;
- **(c)** `-ConnectReferences` posts twice on an OK/WARNING import and sets `ConnectPassStatus`, and the connect pass is **skipped** on an ERROR status.

Plus two guards beyond the suggestion: no connect pass on a dry-run (even with `-ConnectReferences`), and none by default. The whole `Import-NeoIPCMetadata` block (15 tests) passes.

---

Brar:
Both Strand-2 items verified against `19acb94` — resolved, and I ran the suites locally to confirm: **472 passed / 0 failed / 0 skipped** across `MetadataVerify.Tests.ps1` + `Metadata.Tests.ps1` (fully mocked — no live HTTP), and build-docs is green.

**Order-insensitive verifier compare** — `ConvertTo-NeoIPCCanonical` is correct: null-first, recursively key-sorts both `IDictionary` (the package's `[ordered]`) and `PSCustomObject` (DHIS2's read-back) into a canonical `[ordered]`, preserves array order, and passes scalars through. The `-isnot [string]` guard on the `IEnumerable` branch correctly avoids the string-as-char-array trap. Switching to `ConvertTo-Json -InputObject` (from the pipeline form) is the right call here and only collapses the benign empty-array-vs-null case, as you noted. The two tests pin both directions — key-reordered `renderType` is **not** flagged, genuinely-different nested value **is** — so the false positive is gone without masking real drift.

**Untested import behaviours** — the six new tests are well-targeted and match the implementation: `ErrorMessage` joins `message / devMessage` only on a bare WebMessage (no `.response`) and stays null on the happy path; the connect pass fires a second POST only when `-ConnectReferences` and the *response-level* status is OK/WARNING (test (e) correctly exercises the `response.status=ERROR` precedence over a top-level WARNING), and is skipped by default and on `-DryRun`.

That clears every finding on this PR. Nice work — resolving the two remaining threads. 👍

---

Brar:
## Strand 2 addressed — 19acb94

Both Strand 2 findings from the [review](https://github.com/NeoIPC/Surveillance-Toolkit/pull/80#pullrequestreview-4597123384) are fixed and their threads resolved.

**Order-sensitive nested-object compare (`MetadataVerify.ps1`)** — the verifier's `FieldMismatch` branch now canonicalizes both sides before comparing: a recursive `ConvertTo-NeoIPCCanonical` sorts the keys of every nested object (the package's `[ordered]`/`IDictionary` side and DHIS2's `PSCustomObject` read-back alike) and preserves array order. A package object and its read-back holding identical data in a different key order (`renderType { DESKTOP, MOBILE }`, `style { icon, … }`) no longer serialize to different strings, so they stop producing a spurious mismatch that would block a good seed. Scalars and array order pass through, so genuine value drift still surfaces.

**Untested import behaviours (`Import-NeoIPCMetadata`)** — added six tests covering the two commits' headline behaviour: the bare-WebMessage `ErrorMessage` surfacing (joined `message / devMessage` on a persistence-layer failure with no `.response`; null on the happy path) and the `-ConnectReferences` connect pass (a second apply on OK/WARNING with `ConnectPassStatus` set, skipped on ERROR, skipped on a dry-run, off by default).

**Tests** — the full `MetadataVerify` suite (17, +2) and the `Import-NeoIPCMetadata` block (15, +6) pass. An independent adversarial pass over the canonicalizer confirmed it suppresses key-order false positives without masking any genuine scalar / nested-key / array drift; the only old-vs-new outcome change is the benign `package=@()` vs `DHIS2=null` equivalence, which removes noise rather than hiding a drop.

With Strand 1's warning-leak fix (e82cb97) and Strand 2 now both in, all six review threads are resolved.

---

PR #81, opened 2026-07-01
PowerShell approved-verb hygiene + centralized build-report construction

Brar:
Two related pieces of PowerShell housekeeping in `scripts/`.

**Approved-verb hygiene.** Every command script now uses an approved `Get-Verb` verb chosen by behaviour — the report wrappers and the data-dictionary / metadata / protocol builders move from `New-*` / `Make-` to `Build-*` (they render artifacts, not in-memory objects; `New-*` stays for the object-returning module functions), and `New-PartnerReports` becomes the singular `Build-PartnerReport`. The rule is registered as a PowerShell language-tier guardrail and enforced in CI: `PSUseApprovedVerbs` for functions plus a custom Pester check for script *file* names (which `PSUseApprovedVerbs` does not lint).

**Centralized build-report construction.** The five report wrappers each hand-rolled their build-report JSON (an `$extraFields` bag, plus per-step logs with manually-mutated status/exitcode/messages), so five parallel schemas drifted apart. The construction now lives in the `NeoIPC-Tools` module: a schema-owning `Write-NeoIPCBuildReport` (common fields as first-class parameters emitted with fixed camelCase keys) plus `New-`/`Complete-NeoIPCBuildStep` and `Get-NeoIPCParameterSnapshot`. The wrappers declare their inputs through that API rather than assembling hashtables, so the schema is consistent by construction — locale/format harmonized to arrays, uniform per-step logging across the multi-render wrappers, and a reproducibility parameter snapshot on every wrapper. Adds the module's first tests for the build-report shape, and renames the report wrappers' path-typed variables and the affected module parameters to the `...FilePath` / `...DirPath` / `...FileName` framework.


scripts/Build-PartnerCertificate.ps1
```diff
@@ -130,10 +130,16 @@ $quartoVerbosityArgs = switch ($logLevel) {
     default   { @() }
 }
 
+# Snapshot the bound parameters in script scope — inside the Invoke-WithNeoIPCAuth
+# scriptblock $PSBoundParameters is the scriptblock's own (empty) dictionary. Feeds the
+# build report's reproducibility fields.
+$paramSnapshot = Get-NeoIPCParameterSnapshot -BoundParameters $PSBoundParameters
```
Brar, line 136: **Blocker — a live `DirectoryInfo` is serialized into the parameter hash and the JSON build report.**

`-SignatureImagePath` is declared `[System.IO.DirectoryInfo]` (line 6), so `$PSBoundParameters['SignatureImagePath']` holds a live `DirectoryInfo` object at bind time. The reassignment to a relative string at line 99 updates the **variable**, not the `$PSBoundParameters` entry. So here `Get-NeoIPCParameterSnapshot` serializes that object via `ConvertTo-Json -Depth 100 -Compress`, and it's serialized **again** through `Write-NeoIPCBuildReport -Parameters $paramSnapshot.source` (also `-Depth 100`) at line 240.

Consequences:
- `parameterHash` is computed over a huge, machine-specific object graph (`DirectoryInfo` recurses through `.Parent`/`.Root` to the filesystem root), so it is **not reproducible across machines** — defeating the snapshot's purpose.
- `parameters.SignatureImagePath` balloons into a large nested blob in the report JSON.
- Property access on `DirectoryInfo` members for an inaccessible path can throw, tripping the `try/catch` in `Write-NeoIPCBuildReport` and losing the report.

This is exactly the "renamed to a `…Path` variable but the bound value is still an object" hazard. The type is also semantically wrong — the signature is a *file*, not a directory.

**Fix:** type the parameter as `[string]` and resolve inside the body (adjust line 99 to `Resolve-Path -LiteralPath $SignatureImagePath ...`). Alternatively, overwrite the bound entry with the resolved string before snapshotting (`$PSBoundParameters['SignatureImagePath'] = $SignatureImagePath`), but `[string]` is cleaner. The other four wrappers bind only strings/switches/arrays and are clean.

_(Reasoned from source, not executed. A runtime check — render one certificate with `-JsonReport` and inspect `parameters`/`parameterHash` — would give final confirmation.)_

---

.github/workflows/build.yml
```diff
@@ -96,14 +96,24 @@ jobs:
...
+
+            - name: Check PowerShell verb hygiene
+              shell: pwsh
+              run: |
+                # Approved-verb gate: fail the build on any scripts/*.ps1 whose file name uses an unapproved
+                # verb, and (via PSScriptAnalyzer's PSUseApprovedVerbs) any function that does. PSUseApprovedVerbs
+                # lints function definitions only — the file-name half is a custom check in the Pester test.
+                Install-Module -Name Pester -MinimumVersion 5.5 -Force -Scope CurrentUser -SkipPublisherCheck
+                Install-Module -Name PSScriptAnalyzer -Force -Scope CurrentUser -SkipPublisherCheck
+                Invoke-Pester -Path scripts/modules/NeoIPC-BuildTools/Tests/ScriptVerbHygiene.Tests.ps1 -CI
```
Brar: **New test added to the legacy `NeoIPC-BuildTools` module.**

`NeoIPC-BuildTools` is legacy — everything in it is slated to move into `NeoIPC-Tools`, so new code shouldn't land there. The new `ScriptVerbHygiene.Tests.ps1` is a repo-wide gate (it resolves `$repoRoot` and scans `scripts/`; it has no dependency on BuildTools internals), and its sibling `BuildReport.Tests.ps1` was correctly placed under `NeoIPC-Tools/Tests/`. Putting this one under `NeoIPC-BuildTools/Tests/` grows the surface that later has to be migrated and rewired.

**Fix:** move the file to `scripts/modules/NeoIPC-Tools/Tests/ScriptVerbHygiene.Tests.ps1` and update this path — it can just be appended to the existing run on line 99 (`...DataDictionary.Tests.ps1,...BuildReport.Tests.ps1,...ScriptVerbHygiene.Tests.ps1`), dropping this separate step.

---

scripts/Build-PartnerCertificate.ps1
```diff
@@ -220,15 +236,14 @@ finally {
...
-        scriptTimestamp = $scriptTimestamp
-        outputDirPath = $outputDirPath
-        outputLocale = $OutputLocale
-    }
-    $reportPath = if ($JsonReport) { $buildReportFilePath } else { $null }
-    $status = Write-NeoIPCBuildReport -Name 'Partner-Certificate Build' `
-        -Errors $errors -OutputFiles $outputFiles -BuildCompleted $buildCompleted `
-        -StartedAt $startedAt -BuildReportPath $reportPath -ExtraFields $extraFields
+    $reportFilePath = if ($JsonReport) { $buildReportFilePath } else { $null }
+    $status = Write-NeoIPCBuildReport -Name 'Partner-Certificate Build' -StartedAt $startedAt `
```
Brar, line 241: **Minor — per-site builder omits the `siteCodes` summary field.** In Acquire mode this wrapper resolves `$siteCodes` (line 164) and renders per site, but the `Write-NeoIPCBuildReport` call passes no `-SiteCodes`. The two other per-site builders emit it — `Build-PartnerReport` and `Build-ValidationReport` both pass `-SiteCodes`. The per-step `buildSteps` carry `siteCode`, but the top-level summary field is absent here, so the report diverges from its peers per the wrapper-alignment rule.

**Fix:** pass `-SiteCodes $siteCodes` in the Acquire branch (and `$null`/omit in the Pass branch), mirroring `Build-ValidationReport`.

---

scripts/modules/NeoIPC-Tools/Public/BuildReport.ps1
```diff
@@ -0,0 +1,225 @@
...
+    $buildReport = [ordered]@{
+        name            = $Name
+        status          = $status
+        startedAt       = $StartedAt
+        completedAt     = $completedAt
+        outputFilePaths = $sortedOutputFilePaths
+    }
+    if ($PSBoundParameters.ContainsKey('ScriptTimestamp'))   { $buildReport['scriptTimestamp']   = $ScriptTimestamp }
+    if ($PSBoundParameters.ContainsKey('OutputDirPath'))     { $buildReport['outputDirPath']     = $OutputDirPath }
+    if ($PSBoundParameters.ContainsKey('SiteCodes'))         { $buildReport['siteCodes']         = $SiteCodes }
```
Brar, line 76: **Nit — `siteCodes` is assigned unwrapped while `outputLocales`/`outputFormats` (lines 73–74) are wrapped in `@()`.** Not a bug: `$SiteCodes` is `[string[]]`-typed so a single element still serializes as a JSON array, and omitting `@()` is actually *correct* for the `-SiteCodes $null` case (Validation combined mode), where `@($null)` would emit `[null]`. But the inconsistent style invites a future "cleanup" edit that turns it into a bug. Either add a one-line comment noting why `siteCodes` is intentionally unwrapped, or normalize the three consistently (guarding the `$null` case).

---

Brar (commented):
## Review — approved with one blocker to fix first

Solid PR. The verb renames are clean and **complete** — I couldn't find a single dangling reference to any old script/function/param name anywhere in the repo (scripts, CI, `.vscode`, docs, both READMEs, the module manifest, tests), and the approved-verb guardrail landed **identically in both `CLAUDE.md` and `.github/copilot-instructions.md`**. The centralized `Write-NeoIPCBuildReport` / `New`+`Complete-NeoIPCBuildStep` / `Get-NeoIPCParameterSnapshot` API is well-documented (all four carry comment-based help), tested, and correctly placed in `NeoIPC-Tools`. Credentials are kept out of the snapshot (`-Token`/`-Password` excluded; the Reference report's `NEOIPC_BACKUP_PASSWORD` is an env var, never a bound param).

One thing to fix before merge, plus a module-placement miss and a few wrapper-alignment gaps.

### 🔴 Blocker
1. **`Build-PartnerCertificate.ps1` serializes a live `[System.IO.DirectoryInfo]` into the parameter hash + build report** (inline at line 136). `-SignatureImagePath` is typed `DirectoryInfo`; the string reassignment at line 99 doesn't update `$PSBoundParameters`, so the snapshot and `Write-NeoIPCBuildReport -Parameters` both `ConvertTo-Json -Depth 100` a machine-specific object graph → non-reproducible `parameterHash`, bloated JSON, and a possible throw that loses the report. Type it `[string]`.

### 🟠 Major
2. **New verb-hygiene test filed under the legacy `NeoIPC-BuildTools` module** (inline at build.yml line 109). `ScriptVerbHygiene.Tests.ps1` is a repo-wide gate with no BuildTools dependency; move it next to `BuildReport.Tests.ps1` under `NeoIPC-Tools/Tests/`.

### 🟡 Minor / wrapper alignment
3. **`Build-PartnerCertificate` omits the `siteCodes` summary field** though it's a per-site build (inline at line 240) — its two per-site peers emit it.
4. **`Build-PartnerReport` emits no `generatedDataFile` block** though it has a JSON-generation path, whereas Reference and (newly) Patient-Data both do. Partner generates *per-site* JSON, so the shape isn't a drop-in `filePath` scalar — either add a Partner-appropriate summary (array of `{siteCode, filePath}`) or note why it's intentionally excluded. Per the wrapper-alignment "ask before diverging" rule, worth a decision either way.
5. **`Build-PartnerReport`'s Quarto render loop hand-rolls an inline output parser** instead of `Invoke-QuartoRender` + `Complete-NeoIPCBuildStep` like the other wrappers, and classifies a "No problem detected" (NoData) result as an **error** rather than the shared `nodata` status. This is **pre-existing** (it lived in `New-PartnerReports.ps1` and never used `Invoke-QuartoRender`), so not a regression — but it's a missed-alignment opportunity now that the surrounding code moved to the step API. Fine to capture as a follow-up.

### 🔵 Nit
6. **`siteCodes` unwrapped vs. `outputLocales`/`outputFormats` wrapped in `@()`** in `BuildReport.ps1` (inline at line 72) — correct as-is, but the inconsistency invites a future breaking "cleanup".

### ✅ Confirmed with evidence
- **(a)** Zero dangling references to any old name — confirmed by repo-wide `rg`.
- **(b)** Approved-verb guardrail present in **both** `CLAUDE.md` and `.github/copilot-instructions.md` — identical text.
- **(c)** Build-report schema envelope/casing/ordering centrally owned and uniform; per-wrapper field sets differ by design (findings #3/#4 are the exceptions). The PR's claim that **Patient-Data gained the same `generatedDataFile` block as Reference is confirmed**.
- **(d)** Comment-based help present on all four new exported functions.
- **(e)** No live `DirectoryInfo`/`FileInfo` serialized — **denied**: see blocker #1 (Partner-Certificate only; the other four wrappers are clean).

Fix the blocker (and ideally relocate the test), and this is good to merge. This comment is the sign-off — a formal GitHub Approve isn't available on one's own PR.

---

scripts/Build-PartnerCertificate.ps1
Brar: Fixed in c3c1927. Typed `-SignatureImagePath` as `[string]` and dropped the `.FullName` at line 99 (`Resolve-Path -LiteralPath $SignatureImagePath -Relative -RelativeBasePath $reportDirPath`). `$PSBoundParameters['SignatureImagePath']` now holds a plain path string, so `Get-NeoIPCParameterSnapshot` and `Write-NeoIPCBuildReport -Parameters` serialise a string rather than a live `DirectoryInfo` graph — reproducible hash, no JSON bloat, no throw. Agreed the type was semantically wrong too (a signature image is a file, not a directory).

.github/workflows/build.yml
Brar: Fixed in c3c1927. Moved `ScriptVerbHygiene.Tests.ps1` to `NeoIPC-Tools/Tests/` — its `$repoRoot` is `Join-Path $PSScriptRoot '..' '..' '..' '..'`, and the depth from `NeoIPC-Tools/Tests` is identical to `NeoIPC-BuildTools/Tests`, so the test itself is unchanged. Folded it into the existing NeoIPC-Tools Pester run (appended to the `DataDictionary.Tests.ps1,BuildReport.Tests.ps1` invocation and added `Install-Module PSScriptAnalyzer` there for the function-name check), and dropped the separate "Check PowerShell verb hygiene" step.

scripts/Build-PartnerCertificate.ps1
Brar: Fixed in c3c1927. `Build-PartnerCertificate` now passes `-SiteCodes $siteCodes` — initialised to `$null`, set in the Acquire (DepartmentCode) branch, left `$null` in the Pass (HospitalName) branch — mirroring `Build-ValidationReport`'s combined/per-site handling.

scripts/modules/NeoIPC-Tools/Public/BuildReport.ps1
Brar: Documented in c3c1927. Added a comment above the `siteCodes` emit explaining it's intentionally unwrapped — nullable (combined / single-render builds pass `$null`, which must serialise as JSON `null`; `@($null)` would emit `[null]`) and already an array when set (`[string[]]`) — with an explicit "do not normalise to `@()`" note so the inconsistency can't be "cleaned up" into a bug.

Brar:
Thanks for the thorough review. All four inline threads are addressed in c3c1927 and resolved:

- **🔴 Blocker (live `DirectoryInfo` in the parameter hash)** — `-SignatureImagePath` is now `[string]`, so the bound value is a plain path and the snapshot + report serialise a string, not a machine-specific object graph. Reproducible hash, no JSON bloat, no throw. (The type was also semantically wrong — a signature image is a file.)
- **🟠 Major (new test in the legacy `NeoIPC-BuildTools` module)** — `ScriptVerbHygiene.Tests.ps1` moved to `NeoIPC-Tools/Tests/` (identical `$repoRoot` depth, so unchanged) and folded into the NeoIPC-Tools Pester CI step; the separate "Check PowerShell verb hygiene" step is gone.
- **🟡 Minor (Certificate omits `siteCodes`)** — now emitted (`-SiteCodes $siteCodes`, set in Acquire mode, `$null` in Pass mode), matching its per-site peers.
- **🔵 Nit (`siteCodes` unwrapped)** — documented in place with an explicit "do not normalise to `@()`" note, so the inconsistency can't be cleaned up into a bug.

Two body-only findings — decisions:

**#4 — Partner emits no `generatedDataFile`: intentionally excluded** (flagging here per the "ask before diverging" rule). Partner generates JSON **per site**, and each generation is already a first-class `buildSteps` entry (with `outputFileName` + success/error status) *and* its path is in `outputFilePaths`. A scalar `filePath` doesn't fit, and an `{siteCode, filePath}` array would duplicate what `buildSteps` already records — more richly. Partner is the only wrapper with a per-step JSON-gen log, so it captures this *better* than a summary object would; Reference and Patient-Data emit `generatedDataFile` precisely because they have no per-step log for their single JSON.

**#5 — Partner render loop hand-rolls the parser + treats NoData as error**: agreed it's pre-existing (from the old `New-PartnerReports.ps1`) and out of scope for a behaviour-preserving centralization. Captured as a follow-up task (`partner-report-quarto-render-alignment.md`) to move it onto `Invoke-QuartoRender` + `Complete-NeoIPCBuildStep` and map NoData → `nodata`.

Verified locally: the 19 module + verb-hygiene tests pass from the new test location (the function-name `PSUseApprovedVerbs` check runs with PSScriptAnalyzer installed).

Brar:
Verified `c3c1927` against the working tree — all four inline threads check out:

- **🔴 Blocker** — `[string]$SignatureImagePath` (line 6) and line 99 resolves the string directly (`Resolve-Path -LiteralPath $SignatureImagePath …`, no `.FullName`). `$PSBoundParameters['SignatureImagePath']` now holds a plain path, so the snapshot and `Write-NeoIPCBuildReport -Parameters` serialise a string — reproducible hash, no object graph, no throw. ✅
- **🟠 Major** — `ScriptVerbHygiene.Tests.ps1` now lives under `NeoIPC-Tools/Tests/`, folded into the combined `Invoke-Pester` run (build.yml L102); the legacy-module copy and the separate step are gone. ✅
- **🟡 Minor** — cert passes `-SiteCodes $siteCodes` (L246), `$null` in the Pass branch, matching its per-site peers. ✅
- **🔵 Nit** — the "do not normalise `siteCodes` to `@()` — that reintroduces the `[null]` bug" guard comment is in place (BuildReport.ps1 L75). ✅

Both body decisions are sound:
- **#4 (Partner has no `generatedDataFile`)** — agreed it's the right call. Per-site JSON is already a first-class `buildSteps` entry (path also in `outputFilePaths`), which captures more than a summary scalar would; Reference/Patient-Data emit the summary precisely because they lack a per-step log. Good call flagging it per the alignment rule.
- **#5 (Partner Quarto parser / NoData→error)** — correctly scoped out of a behaviour-preserving centralization and captured as a follow-up task.

No open concerns from my side — the review is fully addressed.

---

PR #82, opened 2026-07-02
R logging under knitr, M_neoipcapp app access, and e2e report personas

Brar:
Supports the neoipc-app Playwright e2e suite and hardens report logging.

- **R logging under knitr** — `configure_logging()` no longer installs its global warning/message calling handlers under knitr (they abort a Quarto render); it registers knitr output hooks instead, so render-time warnings and messages reach the unified log channel while staying out of the report body. The report `_setup.qmd` import chunks drop their `warning=FALSE`/`message=FALSE` so the conditions actually reach the hook and get logged.
- **App access + e2e personas** — grants `M_neoipcapp` to the report/admin roles and adds play-report users, so the neoipc-app e2e personas can drive report generation.


reports/common/logging.R
```diff
@@ -32,6 +32,24 @@
...
 
+# Strip knitr's per-line comment marker (the chunk `comment` option, default
+# "##") from the text a knitr warning/message output hook receives, so the log
+# carries the condition text without the rendered comment decoration. `options`
+# is the merged chunk option list knitr passes to the hook (`options$comment` is
+# the active marker). knitr's own "Warning[ in <call>]:" label on warnings is
+# left in place — the call context it carries is useful in the log.
+.knit_condition_text <- function(x, options) {
+  comment <- options$comment
+  lines <- strsplit(x, "\n", fixed = TRUE)[[1]]
```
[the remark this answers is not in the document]
Brar: Checked this against the installed knitr 1.51 source — the payload is a **length-1 character vector** through knitr's entire built-in condition path, so `strsplit(x, "\n", fixed = TRUE)[[1]]` sees every line:

- `sew.message` → `msg_wrap(paste(conditionMessage(x), collapse = ""), …)` — collapsed to one string.
- `sew.warning` → `msg_wrap(sprintf("Warning%s: %s", call, conditionMessage(x)), …)` — one string for any base `warning()`.
- Inside `msg_wrap`, `xfun::str_wrap()` returns a single string with embedded `\n` (verified: `length == 1`), and `comment_out` → `line_prompt` is a `paste0(prompt, gsub("(?<=\n)", continue, x))` — it prefixes lines *within* the string and preserves length 1 (verified by running the pipeline on a multi-line warning: hook `x` length 1).

The only way `x` arrives with length > 1 is a custom condition whose `conditionMessage()` is itself a multi-element vector flowing through `sew.warning`'s vectorized `sprintf` — not reachable via base `warning()`/`message()`, which is all the reports and neoipcr raise. So no text is dropped in practice.

That said, hardening costs one token if we ever want it: `strsplit(paste(x, collapse = "\n"), "\n", fixed = TRUE)[[1]]`. Happy to add it, but as-is the code matches knitr's actual contract. Resolving on that basis unless you'd rather have the defensive collapse.

---

reports/common/logging.R
```diff
@@ -76,13 +94,39 @@ configure_logging <- function(report = NULL, verbosity = NULL) {
...
+  # log channel and the hook returns "" so the condition stays out of the
+  # rendered report body. This only fires for conditions knitr actually surfaces
+  # to a hook — a chunk that sets warning=FALSE / message=FALSE drops the
+  # condition first, so the report _setup.qmd chunks leave those unset where the
+  # log capture is wanted. skip_formatter keeps logger's glue formatter from
+  # trying to interpolate braces in arbitrary condition text.
+  if (!isTRUE(getOption("knitr.in.progress"))) {
+    logger::log_warnings(muffle = FALSE)
+    logger::log_messages()
+  } else if (requireNamespace("knitr", quietly = TRUE)) {
```
Brar, line 122: **Minor — body cleanliness now has a single point of failure: this hook being installed.** Previously the `_setup.qmd` chunks' `warning=FALSE`/`message=FALSE` kept conditions out of the rendered body regardless of logging state; those were (correctly) removed so conditions reach the hook. But that means: a report that forgets `configure_logging()`, calls it after an import chunk, or a future refactor that reorders the setup, now prints raw R warnings into the published PDF/HTML — a leak the old chunk options would have masked. Reachability today is nil (all four setups install the hook before or at the top of their first import chunk, and knitr sews conditions *after* chunk evaluation, so even the same-chunk Partner-Certificate case is covered), so this is latent fragility, not a live bug.

A defensive `warning=FALSE` can't come back (it drops the condition before the hook — the whole point of removing it), so the invariant is purely conventional: **the hook must be installed before any condition-raising code**. Worth a line in the CLAUDE.md Logging section (and its copilot-instructions sibling) so the next report author inherits the rule.

---

reports/common/logging.R
```diff
@@ -76,13 +94,39 @@ configure_logging <- function(report = NULL, verbosity = NULL) {
...
+    knitr::knit_hooks$set(
+      warning = function(x, options) {
+        logWarn(logger::skip_formatter(.knit_condition_text(x, options)))
```
Brar, line 125: **Minor / GDPR note — condition text is now a logged surface.** These hooks route the *full text* of every render-time `warning()`/`message()` into the log channel (`NEOIPC_LOG_FILE`/stderr). To be clear about what this does **not** touch: the DHIS2 query-trace boundary is intact — `log_dhis2_request` (URL + status + row count, never bodies) doesn't flow through R's condition system, and the report body is protected by the `""` returns. But any report or neoipcr condition that interpolates a data-derived value now reaches the operator log. Today that's fine — e.g. Reference-Report's data-protection warning interpolates `n_departments`, an aggregate — but it establishes the pattern. Suggest a sentence in the logging docs (CLAUDE.md Logging section) that render-time condition messages are a logged surface and must carry only aggregates/structural text, never record-level identifiers.

---

metadata/common/userRoles.csv
```diff
@@ -2,11 +2,11 @@ id,code,name,description,authorities,restrictions,sharing
 HhPyparp7LL,,Base,The base role with no explicit right but dashboard and browser cache cleaner access.,M_dhis-web-cache-cleaner M_dhis-web-dashboard,,PUBLIC_RW
 IeOFqOq8C64,,Data entry,A role for users who can enter data via tracker capture,M_dhis-web-tracker-capture,,PUBLIC_RW
 GuDCEG5fXdn,,Metadata export,Gets the right to export metadata,F_METADATA_EXPORT,,PUBLIC_RW
-Kx7Qm2Vb9Lt,NEOIPC_ADMIN,NeoIPC Admin,NeoIPC reporting administration: Reference Report live generation plus reference-data and validation-exception management in the NeoIPC reporting app.,F_NEOIPC_ADMIN,,PUBLIC_RW
+Kx7Qm2Vb9Lt,NEOIPC_ADMIN,NeoIPC Admin,NeoIPC reporting administration: Reference Report live generation plus reference-data and validation-exception management in the NeoIPC reporting app.,F_NEOIPC_ADMIN M_neoipcapp,,PUBLIC_RW
```
Brar, line 5: **Verified correct — `M_neoipcapp` is exactly the authority DHIS2 2.40.3.2 derives, hyphen stripped.** This string looked suspect against the in-file siblings (`M_NeoIPC_Unit_Report` preserves separators; bundled `M_dhis-web-*` keep hyphens), so I checked dhis2-core 2.40.3.2 (`App.java`):

```java
public String getSeeAppAuthority() {
    if (isBundled()) return SEE_APP_AUTHORITY_PREFIX + AppManager.BUNDLED_APP_PREFIX + getShortName();
    return SEE_APP_AUTHORITY_PREFIX
        + getShortName().trim().replaceAll("[^a-zA-Z0-9\\s]", "").replaceAll("\\s+", "_");
}
```

The non-bundled regex strips hyphens (non-alphanumeric, non-space): `neoipc-app` → `neoipcapp` → `M_neoipcapp` ✅. The siblings stay consistent — "NeoIPC Unit Report" has no specials so only spaces→underscores; bundled apps take the no-regex branch, which is why `M_dhis-web-dashboard` keeps its hyphens. And `getKey()` uses the *different* URL-friendly regex `[^A-Za-z0-9\s-]` (hyphen **preserved**), which is why the app correctly serves at `/api/apps/neoipc-app/` while its see-authority has no hyphen — the asymmetry is upstream, not a mistake here. Scoping also checks out: added to exactly `NEOIPC_ADMIN` + `ReportViewer`, nothing else, and the ordinal authority ordering within both rows holds.

---

Brar (commented):
## Review — approved; two doc-level minors, everything substantive verified against upstream source

Both halves check out. The R-logging half was verified against the **installed knitr 1.51 source** (not docs), and the metadata half against **dhis2-core 2.40.3.2** — the two load-bearing upstream contracts in this PR.

### Verified against source
- **The hooks genuinely keep conditions out of the report body.** Both hooks end in `""`, and knitr's `msg_wrap` emits the hook's return value verbatim (`(knit_hooks$get(type))(comment_out(message, options$comment), options)`), so an empty return contributes nothing. Because they *replace* the default hooks, no empty comment-fenced block appears either. This holds for `include: false` import chunks too — output hooks fire during output assembly; `include=FALSE` only discards the final assembled string.
- **The knitr-in-progress guard and the abort rationale are accurate.** `getOption("knitr.in.progress")` is knitr's own sentinel; evaluate wraps chunk code in `withCallingHandlers`, so `globalCallingHandlers` genuinely can't be installed mid-render — the `else if` hook path is the right fix, and `skip_formatter()` correctly stops glue interpolating braces in arbitrary condition text.
- **Ordering is sound in all four reports.** The hook is installed before (or in the same chunk as, and prior to — knitr sews conditions *after* chunk evaluation, covering Partner-Certificate's same-chunk import) every import. Reference-Report was correctly left untouched: its chunks were already in the target state and now benefit from the shared hook.
- **`.knit_condition_text` matches knitr's actual behavior** — knitr applies `comment_out(x, options$comment)` before the hook, prefixing every line uniformly; the per-line strip inverts exactly that, and the preserved `Warning[ in <call>]:` label is confirmed in `sew.warning`'s `sprintf`.
- **`M_neoipcapp` is correct** (inline at `userRoles.csv:5`) — it looked inconsistent with the in-file siblings, but dhis2-core 2.40.3.2's `getSeeAppAuthority()` strips non-alphanumerics before underscoring spaces, so `neoipc-app` → `M_neoipcapp` is exactly what the server derives. The key/authority asymmetry (`/api/apps/neoipc-app/` vs `M_neoipcapp`) is upstream's two different regexes, and the grant is scoped to exactly `NEOIPC_ADMIN` + `ReportViewer`.
- **Copilot's inline on `.knit_condition_text`** — rebutted with evidence in that thread: the hook payload is length-1 through knitr's entire built-in condition path (`paste(collapse)`/`sprintf` → `str_wrap` → `line_prompt` all preserve a single string), so `strsplit(x, …)[[1]]` drops nothing for any base `warning()`/`message()`. A one-token defensive collapse is available if wanted.
- **No credentials or personal names in the personas.** `metadata/play/users.csv` has no password/token column; the added rows are synthetic ("Play", "AT/CH Report User N") and match the e2e definitions; a scan of `metadata/play/` for secrets came back clean.

### 🟡 Minors (inline, both doc-level)
1. **Body cleanliness now has a single point of failure** (inline at `logging.R:119`) — the old chunk-level suppression was the belt-and-suspenders; now the invariant "hook installed before any condition-raising code" is purely conventional. Worth a line in the CLAUDE.md Logging section (+ copilot-instructions sibling).
2. **Condition text is now a logged surface** (inline at `logging.R:122`) — the DHIS2 query-trace boundary and the report body are untouched, but render-time `warning()`/`message()` text now lands in the operator log; the docs should say conditions must carry only aggregates/structural text.

Sub-nit, no change wanted: `.knit_condition_text` would strip a legitimate leading `##` from a condition's own text — affects log cosmetics only, never the body.

No sync obligation triggered (CLAUDE.md/copilot-instructions untouched — though minor #1/#2 propose adding to them). Good to merge with the two doc notes addressed or consciously deferred. This comment is the sign-off — a formal GitHub Approve isn't available on one's own PR.

---

reports/common/logging.R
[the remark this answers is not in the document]
Brar: Verified against knitr 1.51: the warning/message hooks fire once per condition with a length-1 string — a multi-line message arrives as one string with embedded `\n`, not a character vector — so no text is dropped today. Made `.knit_condition_text` flatten with `unlist()` anyway (14571f8), so it stays correct if a hook ever hands over a vector.

reports/common/logging.R
Brar: Documented in the Logging section of `CLAUDE.md` (and the `.github/copilot-instructions.md` sibling) in 14571f8: the render-time hook is the only thing keeping raw conditions out of the rendered body, so `configure_logging()` must run before any condition-raising code. A defensive `warning=FALSE` can't come back (it drops the condition before the hook), so the invariant is conventional — and now written down for the next report author.

reports/common/logging.R
Brar: Documented in the same Logging section (14571f8): render-time condition text is a logged surface and must carry only aggregates/structural text, never record-level identifiers. Also noted that the `log_dhis2_request` boundary (URL + status + row count, never bodies) is separate and unaffected.

metadata/common/userRoles.csv
Brar: Agreed, thanks for the upstream check — `M_neoipcapp` is the non-bundled see-app authority DHIS2 derives (`neoipc-app` → alnum-only regex strips the hyphen → `M_neoipcapp`), scoped to exactly `NEOIPC_ADMIN` + `ReportViewer`. No change needed.

Brar:
Addressed the review in 14571f8:

- **knitr hook payload** — verified the warning/message hooks pass a length-1 string in knitr 1.51 (one call per condition; a multi-line message arrives as one string with embedded newlines), so nothing is dropped today. `.knit_condition_text` now flattens with `unlist()` regardless, so it stays correct if a vector is ever handed over.
- **Logging invariants (docs)** — documented in the Logging section of `CLAUDE.md` + `.github/copilot-instructions.md`: the render-time hook is the sole thing keeping raw conditions out of the rendered body, so `configure_logging()` must run before any condition-raising code; and render-time condition text is a logged surface that must carry only aggregates/structural text, never record-level identifiers (the `log_dhis2_request` boundary is separate and unaffected).
- **`M_neoipcapp` authority** — confirmed correct (the see-app authority DHIS2 derives for the non-bundled app, hyphen stripped), no change needed.

All four inline threads resolved.

Brar:
Verified `14571f8` against source — all four threads check out:

- **knitr hook payload** — `.knit_condition_text` now flattens with `unlist(strsplit(x, "\n", fixed = TRUE))`, and the comment accurately states the contract (length-1 today, defensive flatten). I also verified it *behaviorally*: on the previous head a synthetic multi-element vector dropped everything after the first element; on `14571f8` all lines survive with comment markers stripped. So the code now matches knitr's actual contract *and* is robust beyond it — the right resolution of the Copilot thread. ✅
- **Logging invariants documented** — the new Logging paragraph carries both invariants precisely (hook is the sole body protection → `configure_logging()` before any condition-raising code; condition text is a logged surface → aggregates/structural text only, `log_dhis2_request` boundary separate), and the addition is **byte-identical** in `CLAUDE.md` and `.github/copilot-instructions.md` — sync guardrail satisfied. ✅
- **`M_neoipcapp`** — confirmed, no change needed. ✅

`build-docs` was still running when I checked; assuming it goes green as usual, no open concerns — the review is fully addressed.

---

PR #91, opened 2026-07-17
Report render resilience: backtrace on failure, correct figure N

Brar:
Adds `with_error_trace()` to the shared logging layer and wraps every report's render-time import/compute, so a failure deep in neoipcr surfaces with a full R backtrace on the report's log channel instead of an opaque, location-less error.

Pairs with the neoipcr optional-attribute NA fix (NeoIPC/neoipcr#19): the birth-weight and gestational-age distribution figures now plot only patients with a recorded value, so their sample-size captions report the recorded count rather than the total patient count.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -68,7 +68,7 @@ Reports live under `reports/`:
 
 ### Shared Infrastructure
 
-- **Shared R code**: `reports/common/` — `helpers.R` (locale parsing, string resource loading, DHIS2 connection helpers), `load-neoipcr.R`, `parse-args.R` (CLI arg parsing), `getDataset.R` (dataset export), `logging.R` (unified `logger`-based logging: `configure_logging()` + `logInfo`/`logVerbose`/`logDebug`/`logWarn`/`logError`), `reference.docx` (Word template)
+- **Shared R code**: `reports/common/` — `helpers.R` (locale parsing, string resource loading, DHIS2 connection helpers), `load-neoipcr.R`, `parse-args.R` (CLI arg parsing), `getDataset.R` (dataset export), `logging.R` (unified `logger`-based logging: `configure_logging()` + `logInfo`/`logVerbose`/`logDebug`/`logWarn`/`logError`, plus `with_error_trace()` to log a full backtrace when a render-time computation fails), `reference.docx` (Word template)
 - **Base string resources**: `reports/common.yaml` (English domain terms, table headers, footnotes)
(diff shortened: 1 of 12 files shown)


Brar (commented):
## Review — render resilience (`with_error_trace`) + corrected figure N

**Verdict: correct and well-crafted — approve to merge.** `build-docs` green; Copilot found nothing, and I concur after tracing it and exercising the helper.

### Verified correct
- **`with_error_trace()`** (`reports/common/logging.R`) — I ran it: on error it logs exactly once (message + `rlang::trace_back()`) and then **re-raises** the original error (it doesn't swallow it); on success it returns the value unchanged. The design is right: `withCallingHandlers` runs the handler *before* the stack unwinds, so `trace_back()` captures the failing frame; `expr` is a lazy promise forced inside the handler context; `try(..., silent = TRUE)` ensures a logging-channel failure (e.g. an unwritable `NEOIPC_LOG_FILE`) can never replace the original error; and `logger::skip_formatter` stops glue from interpolating `{}` in the trace text. Good, careful helper.
- **Report wrapping** — the import/compute step in all five `_setup.qmd` (Partner-Certificate, Partner-Report, Patient-Data, Reference, Validation) is wrapped consistently, and the Reference/Partner wrappers correctly include the downstream `|> calculate_reference_data()` / `calculate_department_data()` inside the trace scope.
- **Figure-N fix** — captions now report the *recorded* count (`sum(...$frequency$n)`) instead of the total patient count, correctly pairing with neoipcr #19 (the figures now plot only patients with a recorded value). Confirmed the columns exist: `birth_weight_figure$frequency` carries `dataset` (own/ref, added in neoipcr `calc-api.R`) and `n`, so the Partner `sum(filter(frequency, dataset == "own")$n)` and the `is.null(...) → 0L` guard are valid. Incidentally this also brings the **Partner** captions onto `format_integer()` (they previously passed the raw count to `sprintf`), matching the Reference captions and the digit-group-separator guardrail — `format_integer` is the shared `helpers.R` helper the report sources.
- **Docs** — `CLAUDE.md` and `.github/copilot-instructions.md` got the identical `with_error_trace()` edit (in sync).

### Not verified here (honest scope)
`build-docs` builds the docs, not a live report render (the `_setup.qmd` paths need a DHIS2 connection), so the wrapping and caption changes are validated by inspection + the paired neoipcr #19 tests rather than an executed render. The R semantics of `with_error_trace` are covered by the empirical check above.

This comment is the sign-off — formal Approve isn't available on one's own PR.

---

PR #93, opened 2026-07-22
First-class metadata codes and expression re-key

Brar:
Authors semantic `code`s on every first-class DHIS2 metadata type that historically carried none — program rules, variables, stages, sections, org-unit levels, the tracked-entity type, and program indicators — plus the generated rule/variable families, in one abbreviated NeoIPC vocabulary (`AGENT`, `NCC`/`CC`, `SURV_END`, `ADM`, `IF_*`, `VAL`, `_VR`, …). Program rule actions and the placeholder validation rule are deliberately left code-less: a code folds into a program rule action's `hashCode` over an order-by-less action collection, which can perturb execution order.

With stable codes in place:
- the type-map name-family is factored into DHIS2-aligned identifiable/nameable bases, closing the drift that let `trackedEntityTypes` silently lose `shortName`;
- the gettext msgctxts move from object UID to code (regeneration-stable; existing German option translations preserved);
- all 708 externalised expression files are re-keyed off the rule/indicator code instead of the mutable rule name (the name sanitiser and folder-collision check give way to a path-safety assertion);
- the generators and the Tracker Capture e2e precondition resolve program stages by `NEOIPC_STG_<token>` code rather than a slot-1 data-element anchor.

A directory-level code-constraint lint (≤50 chars / `^[A-Z][A-Z0-9_]*$` / per-type uniqueness over the committed CSVs) now runs in CI, and the metadata test files were added to the CI Pester invocation. Design reference: `docs/metadata-generation-design.md`, `docs/metadata-pipeline-design.md`.

Verified offline (Pester + the CI invocation) and live: the code-bearing metadata imports into DHIS2 without `E4001`, the reporting integration tests pass, and the Tracker Capture e2e passes 281/281 across chromium / firefox / webkit.


Brar (commented):
## Review — first-class metadata codes + expression re-key

Copilot declined this PR (>300 files), so this is a full independent review. Because the diff is dominated by 707 byte-identical (`R100`) file renames, I did not eyeball it — I reduced it to the real surface (28 modified + 1 new file), then **empirically verified the load-bearing invariants** by loading the module in base and head worktrees, running the readers/generators, and running the Pester suite. I also ran a 5-dimension adversarial cross-check (code validity, re-key integrity, translation preservation, type-map refactor, lint/cruft) with independent verification of every finding.

This is excellent, carefully-sequenced work. The 23 commits build the change in a legible order, the type-map base factoring is defensive (`[ordered] + [ordered]` preserves order **and** throws on a duplicate key, so a re-spelling is a load error, not silent drift), and the mis-assigned-stage fix ships with a CI lint that guards against recurrence.

### What I verified (all green)

- **The re-key is content-preserving.** Reading base with base's reader and head with head's reader and comparing every object per-UID (498 program rules, 747 actions, 11 indicators, 1 validation rule): the re-inlined `condition`/`data`/`expression`/`filter`/`leftSide`/`rightSide` values are byte-identical across the whole package **except exactly 8 fields** — which are the intentional stage fixes below. Zero unintended content drift. (The `git` R100 rename *pairing* cross-wires identical-content blobs, e.g. `J4nrp0CYvO8.filter`→CANCELLED but `.expression`→COMPLETED — that is a blob-matching artifact, not a mis-wire; content-by-object-identity is correct.)
- **The 8 semantic changes are correct and deliberate** (commit `3bf1ccc`): 6 "NeoIPC SSI Organism … name has value" rules moved off the BSI stage (`wDQxhhyAkEU`) onto the SSI stage (`cZhr3Wm2fc8`) whose fields they actually read — on BSI their `SENDMESSAGE` could never fire; and the birth-weight-empty warning moved from a `WARNINGONCOMPLETE` on the Admission stage to a program-wide `SHOWWARNING` on the birth-weight *attribute* (which lives on the registration form). The new `MetadataContent.Tests.ps1` lint (name stage-prefix ↔ `programStage`) catches exactly this class, which offline gates miss because a never-firing rule looks identical to one that correctly does nothing. 
- **Codes are valid** across all 8 newly-coded types: `^[A-Z][A-Z0-9_]*$`, ≤50 chars (`programRules` max is 49 — the ≤50 proof at max slot counts is real and load-bearing), unique per type. The single mint point (`Get-NeoIPCGeneratedObjectCode`) means `code == msgctxt` by construction, and the vocabulary replacement is order-safe (longest-phrase-first).
- **All 39 German translations preserved** (34 non-fuzzy + 3 fuzzy-with-text, identical counts base/head) — 0 lost, 0 altered. The ~14-entry total shrink is UID→code context consolidation of *untranslated* source strings.
- **Type-map refactor introduces no drift** beyond the intended `code` additions (and the `shortName` restoration that the past 2.42 E4000 regression had dropped).
- **Dead code fully removed** — `Get-NeoIPCStageByDataElementId` and `ConvertTo-NeoIPCExpressionPathSegment` have zero remaining references.
- **Tests green**: the metadata + data-dictionary Pester suite passes on head (542 passed / 0 failed / 1 skipped), CI `build-docs` is ✅, and the CI wiring correctly adds the three metadata test files.

### One non-blocking finding

**The translation catalogue is stale for program indicators.** `po/metadata.pot` and `po/metadata.de.po` key only 3 of the 11 program indicators by their authored code; the other 8 are still keyed by their *pre-normalization* identity — 6 by bare UID (`ApYgClc8SVA`, `bpsntltaU4y`, `J4nrp0CYvO8`, `Scio2GndATI`, `SFT8PJJoPRZ`, `taydZHWlYCt`) and 2 by the old code (`NEO_SumAllEnrollments_PI`, `NEO_SumAllSurvEnde_PI`).

- **Root cause** is commit ordering: `5ca8c8b` regenerated the catalogue, then `c35458f` normalized the indicator codes afterward, and the catalogue was never regenerated again (`po/metadata.pot` was last touched by `5ca8c8b`). Everything else's msgctxt was re-keyed (e.g. `programRuleVariables A6dAzZkUnJQ`→`NEOIPC_ADM_COMPLETED`); the indicators alone were missed.
- **Impact is benign today**: all 16 stale entries (8 objects × NAME/SHORT_NAME) have empty `msgstr`, so no current translation is affected — but it breaks this PR's own "code == msgctxt by construction" invariant for these objects and will churn the catalogue (and can orphan a just-added translation) on the next regen.
- **Fix is one command**: re-run the localization export against the coded directory. I confirmed it: `Export-NeoIPCMetadataTranslation` on the head `metadata/common` yields all 11 indicators keyed by code (`NEOIPC_ENROLLMENTS_ACTIVE`, `NEOIPC_ENROLLMENTS_CANCELLED`, … `NEOIPC_SURVEILLANCE_ENDS_SUM`). Worth folding into this PR since code-keying the catalogue is precisely its purpose.

*(Also considered and cleared: the base-composition adds a `code` slot to the `indicatorTypes` type-map entry — but there is no `indicatorTypes` data anywhere in the package, so it is inert, and `IndicatorType` legitimately carries a code in DHIS2. Not a defect.)*

Nothing here blocks correctness — the metadata itself is sound and fully tested. My recommendation is to regenerate the catalogue (one commit) before merge so the localization keys match the codes this PR authored. This comment is the sign-off — a formal Approve isn't available on one's own PR.

---

Brar:
Addressed the review's finding — the stale program-indicator translation keys — in c3a457c.

Re-ran `Export-NeoIPCMetadataTranslation` against the coded `metadata/common`, so all 11 program indicators now key by their authored code in both `po/metadata.pot` and `po/metadata.de.po`. The 8 stale entries are gone: the 6 bare-UID keys (`ApYgClc8SVA`, `bpsntltaU4y`, `J4nrp0CYvO8`, `Scio2GndATI`, `SFT8PJJoPRZ`, `taydZHWlYCt`) and the 2 legacy `NEO_*_PI` keys now read `NEOIPC_ENROLLMENTS_*` / `NEOIPC_MISSING_SURVEILLANCE_END` / `NEOIPC_SURVEILLANCE_ENDS_SUM` / `NEOIPC_SURVEILLANCE_END_COMPLETED`.

Root cause was exactly as you diagnosed: `5ca8c8b` regenerated the catalogue, then `c35458f` normalised the indicator codes afterward and it was never regenerated again. This restores `code == msgctxt` for the indicators too.

**Verified:** the diff is confined to `po/` (84/84 balanced — the 8 keys move, none gained or lost); all 39 non-empty German `msgstr` are preserved (0 lost, 0 altered); no other msgctxt changed; and `msgfmt -c` validation passed during the export.

Brar:
Verified `c3a457c` — the finding is resolved. ✅

I re-checked the regenerated catalogue against source (parsing both files by object identity, not the line diff, since re-keying re-sorts the affected entries):

- **All 11 program indicators now key by their authored code** in both `po/metadata.pot` and `po/metadata.de.po`; **zero** residual stale keys (the 6 bare UIDs + 2 legacy `NEO_*_PI` are gone).
- **The change is confined to exactly those 8 objects.** Set-diff of msgctxt values: 16 removed (8 stale × NAME/SHORT_NAME) ↔ 16 added (8 authored codes × NAME/SHORT_NAME), and **no other msgctxt changed**. The raw ±19 line counts are just sort-relocation of the moved entries, not extra edits.
- **No source or translation churn.** The msgid set is byte-identical (0 added, 0 removed), and all 34 non-fuzzy German `msgstr` are preserved (0 lost, 0 altered) — the 3 fuzzy-with-text and total entry count (2820) are unchanged too.
- `msgfmt -c` passes on both (only the benign default `PO-Revision-Date` header warning).

`code == msgctxt` now holds for the indicators as well, so the whole first-class-code surface is consistent. Diff is catalogue-only (outside the tested code paths; CI was green pre-fix and `build-docs` is re-running). No other issues outstanding from my review.

---

PR #99, opened 2026-07-28
Report catalogue translations from Hosted Weblate

Brar:
German, Spanish and French translations contributed through Hosted Weblate, together with the
msgmerge add-on's routine catalogue maintenance.

The deletion count is dominated by reference-comment stripping: the component is configured to omit
`#:` source locations, so this removes roughly 7,200 of them across the nine catalogues. No
translation content is affected by that part of the diff.


Brar:
Superseded: source locations are informative in 45 .Rmd files, so the component now keeps them (po_no_location = false) rather than stripping ~7,200 reference comments. The catalogues have been re-derived; translation counts are unchanged (de 510, es 180, it 140) and the only remaining difference from main is the PO-Revision-Date header.

PR #100, opened 2026-07-28
Protocol documentation catalogue maintenance from Hosted Weblate

Brar:
Catalogue maintenance only — the msgmerge add-on re-wraps the long single-line entries to the
component's configured line width.

No translation content changes: every `msgid` and `msgstr` is byte-identical after unwrapping, which
is why insertions outnumber deletions roughly five to one.


Brar:
Superseded: the diff was entirely line re-wrapping caused by a component line-wrap setting that disagreed with the repository's --wrap-po newlines. The setting now matches, and the component has been re-derived from main, so there is no longer a difference to merge.

PR #102, opened 2026-07-28
DHIS2 metadata catalogue translations from Hosted Weblate

Brar:
German translations for the DHIS2 metadata catalogue, contributed through Hosted Weblate, together
with the msgmerge add-on's routine catalogue maintenance.

Most of the diff is line re-wrapping rather than translation change, which also puts it past the size
at which automated review is available.


Brar:
**Hold until #106 merges.** This strips 16,354 duplicated `priority:NNN` lines from the six catalogues, and until #106 lands the metadata exporter writes them straight back on its next run — so merging this first buys nothing and reverts silently.

Once #106 is in: merge this with **"Rebase and merge"**, never squash. A squash gives Weblate's commits a different patch identity, so it cannot recognise them as already merged, replays them, and conflicts across every component at once.

For the record on why the flags were duplicated at all: `priority:NNN` is a source-string property. Weblate reads it from `po/metadata.pot` and strips it when writing each language, so the exporter emitting it into every `.po` produced churn Weblate undid on its next write. #106 makes the exporter emit the template alone, which is what makes this strip stick.

PR #103, opened 2026-07-28
Make Weblate the sole writer of the translation catalogues

Brar:
Two writers on one gettext catalogue is what kept breaking every Weblate component at once. po4a
re-generates each `.po` from the template on every run and rewrites `POT-Creation-Date`; Weblate
writes the revision, translator and generator headers into the same hunk. Git cannot auto-merge that,
so a single source-string change conflicted every language of a catalogue simultaneously.

This makes the ownership explicit and enforces it. The repository owns the `.pot` templates; Weblate
owns the `.po` translations and is their only writer. `Invoke-Localization.ps1 -Update` now restores
every Weblate-owned catalogue from `HEAD` after po4a has run, and refuses to start when one is already
dirty. Three catalogues stay repository-owned and are named as such, so "who may edit this file" has
one answer per file rather than a convention people have to remember. A continuous-integration gate
rejects hand-written changes to the Weblate-owned ones.

The antibiotic catalogue moves in the other direction, back to the repository: its content is
CC BY-NC-SA 3.0 IGO, and a NonCommercial term is not a free licence, which Hosted Weblate's free plan
requires. The ATC group descriptions are dropped along the way — only half the groups ever had one,
they restated the group name, and they were the sharpest edge of that licensing question.

Underneath sits a text-file contract, because the ownership rule only holds if every writer agrees on
what the file should look like. A NeoIPC catalogue is written by po4a under WSL, merged by msgmerge,
rewritten by Weblate, uploaded to by Python, and read by R and PowerShell — each with different
newline and encoding defaults. Two catalogues had already reached Weblate as invalid gettext, CRLF
throughout except a few LF-only lines, because one writer produced CRLF and another rewrote only the
header block. Every writer is now pinned individually and the gate inspects the working tree rather
than the index, since `text=auto` hides exactly this defect at commit time.

Finally, the Weblate side of the contract is written down. The four gettext components had accreted
configuration differences nobody had chosen — two line-wrap settings, two source-location settings and
a set of dormant message templates — which went unnoticed because nothing compared them, and one of
which silently re-wrapped a whole catalogue on every write. `docs/weblate-component-configuration.md`
records the canonical settings, a reason for each remaining difference, and which artifact holds which
kind of fact: translations in the `.po`, source-string flags in the `.pot` because that file backs the
source translation, and approval and labels only in Weblate's own store.


Brar (commented):
## Review — Weblate single-writer + text-file contract

Copilot declined again (>300 files), so this is a full solo review: five adversarial dimensions over the checkout with independent verification of every finding, plus my own checks on the parts I care most about. All three CI gates are green.

This is a strong diagnosis. Naming an owner per file, rather than relying on a convention people have to remember, is the right fix for the failure we hit on #90 — and the text-file contract underneath it is the part most projects never get around to. The commit messages are unusually good: they record what was tried and rejected, which is what made verification possible.

### Verified clean

- **The 707 expression files are mechanically safe.** Every expression value is byte-identical after `rstrip` — **zero** value changes. 706 gained a trailing newline; one lost a stray extra blank line (`NEOIPC_SURVEILLANCE_ENDS_SUM`, whose value `d2:condition('#{DDLi3OGC5Yh.Xn8q3giOGcB} > 0',1,0)` is unchanged). Every head file now ends with exactly one newline.
- **No translation lost.** Across all 16 changed catalogues: **0 lost, 0 altered**. The AWaRe rescue *adds* translations (`glossary.es` 0→3, `glossary.it` 0→2), and `antibiotics.de/es` hold at 62/170 through the ATC removal — confirming none of the dropped strings had been translated.
- **The AWaRe move landed exactly right.** `es → "Precaución"`, `it → ""` (deliberately blank), the rationale travelling as a translator note in every catalogue, one `msgctxt "watch"` per file, and the strings genuinely gone from `reports/common.yaml`. This is the outcome I'd hoped for on #90 — the term decided once, with its reasoning attached, rather than re-decided per string.
- **Guardrail files are in sync**: 97 bullets each, zero differences. The asymmetric diffstat is just different starting points.

### Findings

**1. `Invoke-Localization.ps1` — both git guards ignore the exit status, and one failure mode deletes catalogues** *(high)*

`Assert-CleanWeblatePo` treats an empty `git status` result as "clean", and `Restore-WeblateOwnedPo` derives `$inHead` from `git ls-tree` without checking it succeeded. If `ls-tree` fails, `$inHead` is empty, so **every** declared catalogue falls into `$created` — the delete branch — and the run removes them all while the clean-tree guard that should have stopped it has already failed open for the same reason. The `git restore` call twenty lines below *does* check `$LASTEXITCODE`, which is what makes this read as an oversight rather than a decision. Independently confirmed: pwsh 7.6 reports `$PSNativeCommandUseErrorActionPreference = False`, and nothing in the tree opts in, so native failures are silent. The files are committed and recoverable, so this is not data loss — but it is the one guard whose whole purpose is to fail closed. Checking `$LASTEXITCODE` after both reads fixes it.

**2. `infectious-agents/LICENSE.md` declares CC BY 4.0 for catalogues that carry a CC BY-NC-ND 4.0 notice** *(medium)* — all nine `po/infectious_agents.<lang>.po` headers still say NC-ND. Given this PR's own argument is that a NonCommercial/NoDerivatives term cannot apply to a catalogue whose purpose is to be translated, the file contents contradict the licence the PR sets.

**3. The stated reason for dropping the ATC descriptions doesn't match the descriptions** *(medium)* — "each restated the group name" is contradicted by the 17 strings actually removed. Two dimensions found this independently. The *decision* looks right for the licensing reason given; it's the justification sentence in `LICENSE.md`/`README.md` that overstates. Related: commit `3886e9c` says "twenty strings" where 17 were removed (verification judged the arithmetic itself non-actionable, but the two statements should agree).

**4. `scripts/test-update-glossary-po.py` never exercises `update-glossary-po.py`** *(medium)* — I checked this directly: it imports only `os`, `sys`, `tempfile`, `pathlib`, `polib`, `ruamel.yaml`. It tests that *the libraries* write LF, and on Linux even skips its end-to-end control (`os.linesep is LF here, so the end-to-end writer control is skipped`). It would pass with the fix reverted, and the translator-note bug the same commit fixes has no test at all. The name implies coverage it doesn't provide.

**5. Documentation claims that the branch's own later commits invalidated** *(medium, several)* — the guardrails still say `update-glossary-po.py` "writes nothing else" (commit `5dbafe1` restored the `.po` merge); both instruction files still instruct bare `po4a <config>` runs that rewrite nine Weblate-owned catalogues; the gate's remedy for `po/metadata.*.po` is a no-op in the only state the gate observes; and `Test-PoPlaceholders.ps1`'s recorded cause is wrong — the backslash glob works on Linux, the real failure was cwd-relative resolution. Each is a sentence that was true when written and stopped being true two commits later.

**6. Smaller** *(low)*: `Test-TextFileHygiene` skips paths git quotes (non-ASCII filenames) and reports a lone-CR file with a diagnostic asserting a NUL it doesn't contain; the text-hygiene CI job can't catch the working-tree-CRLF case its comment credits it with (checkout normalises first — it's a real local guard, just not a CI one); `PSScriptAnalyzerSettings.psd1` states the wrong `#Requires` floor; a stale comment in `Export-NeoIPCAntibioticTranslation` still says Weblate creates new antibiotic locales.

Three further findings were **refuted** on verification, including the claim that the antibiotic PO writer deleting dropped msgids is a defect — the behaviour predates this PR and caused no loss.

### Recommendation

Only finding 1 is a code defect, and it's a two-line fix. Everything else is documentation that drifted *within the branch* — which is unsurprising at ten commits of this density, and is exactly what this repo's own "re-read iteratively-edited documents end-to-end" guardrail exists for. I'd fix 1, reconcile the licence in 2, and do one pass over the claims in 3–5 before merge. This comment is the sign-off — a formal Approve isn't available on one's own PR.

---

Brar:
Addendum to finding 1 — the verification pass reproduced it end-to-end, and the trigger is more everyday than my write-up implied.

I described the mechanism from reading the code. The reproduction used git's **`detected dubious ownership`** error — the ordinary container / root-over-a-user-checkout case — against a real repo with three committed catalogues, one carrying an uncommitted edit:

```
$ git -C .../owned status --porcelain
fatal: detected dubious ownership in repository at '.../owned'
   exit=128, stdout empty

$ pwsh -File probe.ps1          # verbatim Assert-CleanWeblatePo + Restore-WeblateOwnedPo
Assert-CleanWeblatePo: PASSED (no dirt reported)
  removed po/reports.af.po (po4a created it; Weblate adds new languages from the .pot)
  removed po/reports.de.po (po4a created it; Weblate adds new languages from the .pot)
  removed po/reports.el.po (po4a created it; Weblate adds new languages from the .pot)
EXIT=0
$ ls po/   ->  empty
```

The run ends `Localization update complete.` and exits 0, printing a per-file message telling the operator the deletion was correct. A non-git source tree (`fatal: not a git repository`) gives the same outcome, where the loss is total.

Three details that sharpen it:

- **Blast radius is 27 files** — the three Weblate-owned po4a configs each declare `af de el es et fr it ne tr`.
- **`$LASTEXITCODE` is never reached.** With `ls-tree` failing, `$tracked` is empty, so the one existing check (guarding `git restore`) is skipped entirely.
- **Both functions are new in this PR** — `git show 2b13ce0:scripts/Invoke-Localization.ps1` contains no git invocation at all, so this isn't inherited. And no commit message mentions exit codes or fail-open for these guards, so it reads as oversight against the doc comment's own stated intent.

CI is unaffected (`build.yml` only renders with `--no-update`), which means this lands on developer machines — precisely where uncommitted translator edits live. Still a two-line fix at the `status` and `ls-tree` calls.

One correction to my own review while I'm here: I called the committed content "recoverable", which holds in a real repo once `safe.directory` is fixed, but not in a non-git tree. The uncommitted work the guard exists to protect is destroyed either way.

---

Brar:
All six findings addressed, in `5ffb0a2` and `21ac91b`.

**1 — the guards (fixed in `5ffb0a2`).** Both `git status` and `git ls-tree` now check `$LASTEXITCODE` and throw. Your addendum's reproduction is the decisive part: `dubious ownership` is an everyday trigger, and I had not appreciated that `$tracked` being empty means the one existing exit-code check is never reached. The correction you made to your own review is the one that matters — in a non-git tree the loss is total, and the uncommitted work the guard exists to protect is destroyed either way.

Two things came out of re-reading it with that in hand. The per-file message claimed *"po4a created it"*, which the code cannot know; all that branch establishes is that HEAD does not contain the file, and asserting a cause told the operator the deletion was correct in exactly the case where the partition was wrong. It now states the observation. And the message is 27 files' worth of reassurance, as you measured.

**4 — the glossary test.** You were right that it tests the libraries rather than the script, and right that it would pass with the fix reverted. It now runs `update-glossary-po.py` end to end and checks the bytes of all three artifacts, plus the content round-trip that had no test at all — keys becoming `msgctxt`, a key comment becoming a translator note, and the header block *not* becoming one.

The part worth flagging: byte assertions cannot catch this regression where `os.linesep` is already LF, which is where the test usually runs. So the pinning is also asserted against the source. Verified in both directions — removing the four pins turns it red through the bytes *and* the source scan, restoring them turns it green. Two of my own attempts at that verification were themselves broken (a paren-naive regex that matched `.save(str(pot_path)` and stopped, and a heredoc that mangled the escapes so the mutation silently did nothing and the test "passed"), which is a decent illustration of your point that a test that cannot go red is worse than none.

**3 — the ATC justification.** Confirmed by pulling the removed column out of `3886e9c^`: they described mechanism of action, spectrum and composition, not the group name. The decision stands on the licensing argument — they were the largest block of reproduced WHOCC prose, which is the opposite of restatement. Corrected in `LICENSE.md` and both places in `README.md`. The commit message's "twenty" is in already-pushed history, so I have left it rather than force-push under review.

**5 — the drifted claims.** All four fixed: the guardrail describing the script as writing nothing else, the bare `po4a <config>` examples (now carrying an explicit warning and using the one repository-owned config), the gate's `--source=HEAD` remedy (which by the time that gate fires restores the version being rejected — now `origin/main`), and `Test-PoPlaceholders`'s recorded cause, which was cwd-relative resolution rather than the backslash.

**6 — the small ones.** The analyzer settings' 7.0 floor and the stale "Weblate creates new locales" comment are corrected.

**2 — recorded, not fixed, and deliberately so.** The template says CC BY 4.0 while the locale files still say MIT and CC BY-NC-ND. It cannot be corrected from this side: those files are Weblate-owned, and `msgmerge` preserves a target's header rather than taking the template's, so nothing here will ever update them — a hand-edit would be the two-writer state this PR removes, and the gate would reject it. The route is Weblate. That constraint is now written at `catalogHeaderMap`, where someone will look for it.

Two of the low findings are left as you filed them: `Test-TextFileHygiene` skipping git-quoted paths, and the text-hygiene CI job's comment crediting it with a check that checkout normalisation prevents. Both are real; neither is a merge blocker, and both want a change to what the gate does rather than to what it says.

---

Brar:
Verified `5ffb0a2` and `21ac91b` against source. All fixes land, and the one push-back is correct.

**1 — the guards.** Both `git status` (l.381) and `git ls-tree` (l.415) now check `$LASTEXITCODE` and throw. I re-ran my own reproduction against the fixed code — same dubious-ownership repo, three committed catalogues, one carrying an uncommitted edit:

```
git status exit in this repo: 128
RESTORE THREW (good): Cannot list the Weblate-owned .po files in HEAD
  (git ls-tree exit code 128); refusing to partition them for restore/delete.
--- catalogues still present? --- reports.af.po reports.de.po reports.el.po
  uncommitted edit preserved: 1
```

Fails closed, with a message that names the cause. Good catch on the per-file text too — *"po4a created it"* asserted a cause the code cannot know, and it was doing that in exactly the case where the partition was wrong.

**4 — the glossary test.** Now genuinely a regression guard. It runs the script via `subprocess` and checks all three artifacts, and the source scan covers the four real writer pins (`pot.save`, two `po.save`, one `open(..., "w")`). I confirmed it goes red — removing the pin from `pot.save(str(pot_path))` yields `FAIL … does not pin newline="\n" -- it will emit CRLF on Windows`, **exit code 1** (unmutated: 0), so it can gate.

Worth recording: my *first* attempt at that mutation was invalid — I stripped the first `newline="\n"` in the file, which is a **comment**, and the test correctly stayed green because `_code_only()` deliberately strips comments before scanning. So the false-failure mode your docstring calls out is real and handled. And the `if not saves and not writes` branch is the right guard against the stale-scan failure you hit twice; it is exactly the "test that cannot go red" problem, caught in the test itself.

**3, 5, 6.** All confirmed: `LICENSE.md` now says the descriptions "were also reproduced verbatim, and are **no longer carried at all**" — the restatement claim is gone; the analyzer floor reads 7.6 with the note that it is "measured, not assumed"; the stale "Weblate creates new locales" comment is gone; and the gate now says outright that `--source=HEAD would restore the very version being rejected` before giving `--source=origin/main`. Leaving "twenty" in already-pushed history rather than force-pushing under review is the right call.

**2 — the push-back is correct, and I checked the mechanism rather than taking it.** msgmerge does preserve the target's header:

```
$ msgmerge -o out.po t.po r.pot     # t.po: "OLD LICENCE NOTICE MIT", r.pot: "NEW … CC-BY-4.0"
$ head -1 out.po
# OLD LICENCE NOTICE MIT
```

So the template's CC BY 4.0 will never propagate into the existing locale files, a hand-edit would be precisely the two-writer state this PR removes, and the gate would reject it. Recording the constraint at `catalogHeaderMap` is the right place — that is where someone will be when the question occurs to them. Agreed it belongs on the Weblate side; my finding should have said "resolve via Weblate", not "fix here".

Leaving the two low ones as filed is reasonable — both want a change to what the gate *does*, not to what it says, and neither blocks.

`text-hygiene` and `catalogue-ownership` are green on the new head; `build-docs` is still running. Nothing outstanding from my side once it lands.

---

Brar:
One follow-up from the sibling reviews, landed here as `61bbd72`.

The `[*.{po,pot}]` EditorConfig comment claimed the catalogues "must round-trip byte-for-byte; leave every byte alone". That denies an inheritance which is deliberate: `end_of_line = lf` and `charset = utf-8` still cascade from `[*]`, and should — a catalogue's own header declares `charset=UTF-8`, and LF is the contract this branch establishes. Only `insert_final_newline` and `trim_trailing_whitespace` are suppressed.

The settings are unchanged; the wording now says what they do. It matters here more than elsewhere, because the original sentence invited exactly the wrong fix — overriding `end_of_line` and `charset` for `.po` to make the comment true, which would carve the catalogues out of the contract and reopen the mixed-CRLF corruption this repository has already seen reach Weblate.

The identical stanza is reworded in neoipc-app (`7149fea`) and neoipcr (`a472142`) in the same pass so the three stay in step. I also confirmed the remaining four repositories carry an `.editorconfig` but no gettext catalogues, so the stanza is correctly absent there rather than missing.

---

PR #105, opened 2026-07-29
Give the toolkit a README that orients a newcomer, and state the metadata package's real limits

Brar:
Give the toolkit a README that orients a newcomer, and state the metadata package's real limits

The landing page was one sentence and a licensing table — the first thing anyone
sees on GitHub, and it explained nothing. It now says what the repository is and
who it is for, what each directory holds, how the five products are versioned and
released, and what the tooling needs, linking out to the deeper documents rather
than restating them. Contribution and translation routes are named, since the
protocol and report text are translated by volunteers who never touch git.

The metadata package is now described honestly rather than optimistically. It is
alpha, it does not yet follow the WHO packaging conventions, and it is verified
only against the 2.40 and 2.41 lines; on 2.42 and 2.43 an import intermittently
drops the members of owned ordered collections while reporting success.

Its import instructions also gain something they were missing entirely: the
package must be applied twice. DHIS2 does not link an object's owned reference
collections to objects created in the same payload, so a single import leaves
option-group sets and program-rule actions empty and still reports status=OK.
Following the previous instructions produced a silently incomplete instance.

The declared DHIS2 version in the generated manifest follows what is actually
verified rather than a release known to be defective.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 The NeoIPC Project Consortium
+Copyright (c) The NeoIPC Project Consortium
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -1,5 +1,122 @@
-# Surveillance-Toolkit
-This repository contains the surveillance toolkit of the NeoIPC Project
+# NeoIPC Surveillance Toolkit
+
+[![Build](https://github.com/NeoIPC/Surveillance-Toolkit/actions/workflows/build.yml/badge.svg)](https://github.com/NeoIPC/Surveillance-Toolkit/actions/workflows/build.yml)
+[![Translation status](https://hosted.weblate.org/widgets/neoipc/-/svg-badge.svg)](https://hosted.weblate.org/engage/neoipc/)
+[![License: MIT](https://img.shields.io/badge/license-MIT-blue.svg)](LICENSE)
+
+Everything a neonatal department needs in order to take part in NeoIPC surveillance, and everything
+the project needs in order to analyse what they collect: the surveillance protocol and its case
+definitions, the DHIS2 configuration that gathers the data, the reference lists of infectious agents
+and antimicrobial substances, and the reports that give each participating department its results.
+
+[NeoIPC](https://neoipc.org) works to reduce the transmission of resistant bacteria in neonatal
+intensive care across Europe and globally. Its
+[surveillance system](https://neoipc.org/surveillance/) collects healthcare-associated infection and
+antimicrobial-use data from neonatal departments so that a department can compare itself against the
+network. This repository is the **normative source** for that surveillance: when a case definition
+changes here, the change propagates to the data-collection forms, the analysis code and the reports.
+Where code and the definitions in this repository disagree, the definitions win.
+
+## What's in here
+
+| Path | Contents |
+|------|----------|
+| [`doc/protocol/`](doc/protocol/) | The **NeoIPC Core Surveillance Protocol** in AsciiDoc, including the eight normative case definitions under [`definitions/`](doc/protocol/definitions/) — clinical sepsis, the two laboratory-confirmed bloodstream-infection variants, the three surgical-site-infection depths, necrotising enterocolitis and pneumonia. |
+| [`metadata/`](metadata/) | The **canonical DHIS2 configuration** for the `NEOIPC_CORE` tracker program — data elements, option sets, program rules, tracked-entity attributes and the organisation-unit scaffold — authored as reviewable per-type CSVs plus externalised expression files rather than one opaque JSON blob. See [`metadata/common/README.md`](metadata/common/README.md). |
+| [`metadata/common/infectious-agents/`](metadata/common/infectious-agents/) | The **NeoIPC Infectious Agent List** — a pragmatic ontology of organisms with their synonyms and phenotypic resistance categories, named from LPSN, MycoBank and ICTV. Licensed separately; see below. |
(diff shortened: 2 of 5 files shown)


Brar (commented):
## Review — README + honest metadata-package limits

Reviewed with the three sibling PRs on this branch. Copilot raised nothing here. Three adversarial dimensions with independent verification; most findings were refuted, which is the expected outcome for a PR whose whole purpose is replacing optimistic claims with accurate ones.

The substance is right, and the twice-apply discovery is the valuable part: DHIS2 not linking owned reference collections to objects created in the same payload, while still reporting `status=OK`, means the previous instructions produced a silently incomplete instance. Documenting that — and narrowing the declared version to what is actually verified rather than to a release known to be defective — is exactly the kind of correction that only gets made by someone who went and looked.

### One finding worth fixing before merge

**`Import-NeoIPCMetadataPackage` does not exist** *(medium)* — `metadata/dist/README.md:77` tells importers to run it, and it is the only reference to that name anywhere in the tree. Verified by loading the module rather than by grep alone:

```
Import-Module scripts/modules/NeoIPC-Tools -Force
Get-Command Import-NeoIPCMetadataPackage   # CommandNotFoundException
Get-Command Import-NeoIPCMetadata          # resolves
```

The real cmdlet is **`Import-NeoIPCMetadata`**. This matters more than a typo usually would for two reasons: it sits in the *distributed* package README that ships to importers, and it is the command carrying the new twice-apply instruction — so the one procedure this PR exists to document is the one a newcomer cannot run. Two independent reviewers found it separately.

### Smaller, non-blocking

- The newly-documented import command is written with **no host argument**, so it defaults to the production host. Worth an explicit `-Hostname` in the example, since the surrounding text is aimed at someone standing up their own instance.
- `RELEASING.md`, which the README presents as "the full mechanics", points readers at a `ToDo.md` that isn't in this repository (it lives in the private orchestration workspace) — the same class of dangling reference the sibling app PR removes from its own README.

### Cross-repo consistency

I checked the version claims across all four PRs, since they were written together and each asserts DHIS2 facts. They agree: this PR's "verified on 2.40/2.41, 2.42+ drops owned-collection members while reporting success" lines up with neoipcr#22 narrowing its declared list to `2.40.12.0`/`2.41.9.0` and dropping `2.42.5.1`/`2.43.0.1`, and with app#24's "does not work on 2.42+". No contradiction between the four.

Nothing else outstanding. This comment is the sign-off — a formal Approve isn't available on one's own PR.

---

Brar:
Correcting my own framing on the `Import-NeoIPCMetadataPackage` finding — the defect stands, the consequence I attached to it does not.

I wrote that "the one procedure this PR exists to document is the one a newcomer cannot run." That overstates it. The prose two lines above the broken command already gives the complete remedy — *"**Apply the package twice.** This is not optional and not a retry… The second, identical import runs when every referenced object already exists, and connects them."* So someone who hits `CommandNotFoundException` is **not** nudged toward the single-import failure the section warns about; the instruction survives without the command name, and recovery is one `Get-Command -Module NeoIPC-Tools '*Metadata*'` away.

The finding itself is unchanged and still worth fixing: `metadata/dist/README.md:77` names a command that exists nowhere in the tree, and the fix is one word — `Import-NeoIPCMetadata -ConnectReferences`. The `-ConnectReferences` switch is right, and `Test-NeoIPCMetadataImport` in the same sentence is real, which is what makes the wrong sibling read as authoritative. But it is **low** severity rather than medium, and I should have said so.

Nothing else in that review changes.

---

Brar:
## Review findings addressed

**Fixed — `Import-NeoIPCMetadataPackage` does not exist.** Confirmed by loading the module rather than by grep: that name resolves to nothing, and the real cmdlet is `Import-NeoIPCMetadata`. This was the worst kind of error for this particular change — the invented name sat in the README that ships to importers, on the one instruction the change exists to add, so the twice-apply procedure was the one thing a newcomer could not run.

**Fixed — no explicit host in the example.** The call now names `-Hostname` rather than falling back to a default, which matters because the surrounding text addresses someone standing up their own instance. Verified `-Path`, `-Hostname` and `-ConnectReferences` are all real parameters of the cmdlet's `Path` parameter set before writing the example.

**Fixed — `RELEASING.md` pointed at a task index that is not in this repository.** It lives in the private orchestration workspace, so the reference was dead for anyone reading this repository on its own. Same class of dangling reference the sibling app change removes from its README; the sentence now stands on its own.

Copilot reviewed 4 of 4 files and generated no comments.

### What did not change

Nothing else was raised. The version claims here remain consistent with the sibling changes — verified on 2.40/2.41, 2.42+ dropping owned-collection members while reporting success — which matches neoipcr narrowing its declared list and the app documenting that 2.42+ does not work.

Brar:
Verified `56893a2` — all three land.

| Check | Result |
|---|---|
| Command name | `Import-NeoIPCMetadata -Hostname dhis2.example.org -Path <package>.json -ConnectReferences` |
| Stale name anywhere in the tree | **0 hits** |
| `ToDo.md` in `RELEASING.md` | gone |

I re-verified the parameters the way the original defect was found — by loading the module rather than by grep, since grep is what would have let the invented name through in the first place:

```
Import-Module scripts/modules/NeoIPC-Tools -Force
Get-Command Import-NeoIPCMetadata   -> resolves
  -Hostname            present
  -Path                present
  -ConnectReferences   present
  Path parameter set exists: True
```

So the example is runnable as written, and adding the explicit `-Hostname` closes the quieter of the two problems — an importer standing up their own instance would otherwise have had the example silently default to production.

One note on your framing: you described this as "the twice-apply procedure was the one thing a newcomer could not run." I'd retracted that characterisation shortly before, and I still think the weaker version is the accurate one — the prose immediately above (*"Apply the package twice. This is not optional and not a retry…"*) carries the complete instruction without the command, so the reader who hits `CommandNotFoundException` isn't pushed toward the single-import failure the section warns about. The defect was worth fixing exactly as you fixed it; it just wasn't load-bearing for the procedure. Not asking for a change — only flagging it so the severity doesn't get carried forward into the release notes.

Nothing outstanding here.

---

PR #118, opened 2026-07-30
Hand the glossary catalogues to Weblate

Brar:
Hands `po/glossary.*.po` to Weblate and cuts the generator back to the template, in one change — because
the two halves cannot be separated. Register the component first and the repository is a second writer;
remove the merge first and nothing keeps the catalogues in step with the template at all.

### What the generator loses

Roughly 300 lines whose only caller was the removed merge: the language-name and plural-rule tables, the
contributor filter, the flag merge, and the header/metadata builders. Weblate writes those headers now,
its msgmerge add-on brings the catalogues up to a changed template, and its `new_base` creates one for a
language it adds. `--generate-yaml` is untouched — reading the catalogues is still this script's job.

`scripts/modules/NeoIPC-Tools/Tests/PoHeader.Tests.ps1` is what holds the result to the contract. It
asserts against the **committed** files rather than the writers, so a header that drifts by any route is
caught, and it now runs in CI.

### The test asserts the invariant, not the implementation

It used to check that the script *wrote* catalogues. It now plants a catalogue carrying a stale entry and
a translation, runs the script, and asserts the file is **byte-identical** afterwards.

Both invocations are bracketed, and that is not redundant. Only `--generate-yaml` opens a catalogue into a
mutable object, and it is the mode `Invoke-Localization.ps1` actually runs — a writer planted there passed
the entire suite green while only the bare run was asserted. Each direction is proven rather than assumed:
planted, caught, removed, green.

The CI ownership gate gains `po/glossary\..*\.po`, so the same rule is enforced from the commit side, and
its remediation text gains the glossary case it had no entry for.

### Corrections

Fourteen statements described the retired arrangement. Two were never true rather than merely stale: the
shared exclusion list claimed an R consumer that does not exist, in the future tense; and the text-hygiene
guardrail named this generator as a writer of the infectious-agent catalogue, which it never was.

`.gitattributes` was the last place in the repository saying a hand edit to a glossary catalogue was
legitimate, and it was missed on the first pass because the sweep covered only Markdown, PowerShell,
Python and YAML.

`po/glossary.pot` was also the only wrapped template here — polib defaults to `wrapwidth=78`, giving a
longest line of 91 against 1830 in the po4a-generated ones, while every component declares
`po_line_wrap: 65535`. Now pinned.

### Registration

Not done yet, deliberately: it belongs at merge time, in the same window as this landing. Expect one
flags-only churn commit on the first drain — the retired merge had copied the template's source flags
(`terminology`, `read-only`) into every catalogue and Weblate strips source flags when it writes a `.po`.
`ignore-same` is the exception: it is translator-managed, and German carries two more than the others,
which is real state.


Brar (commented):
## Review — the glossary ownership handover

Copilot reviewed 13 of 13 files and generated **no comments**, so there is nothing to adjudicate and this is the whole review. Note also that `2d0a615` is one commit past what Copilot saw (it reviewed `cfb8013`), so the head commit has had no other reader.

### Verified, by running it rather than reading it

- **The template is byte-reproducible.** Regenerating `po/glossary.pot` from `glossary.yaml` in a clean tree differs only in `POT-Creation-Date`. LF-only, no BOM, longest line 147, and the run writes nothing else under `po/`. The `wrapwidth` pin does what it claims.
- **The CI gate needs no new exemption.** It checks `%ce` per commit against `hosted@weblate.org`, so glossary drain pull requests pass on the same mechanism as the other four. No deadlock, no `auto_lock_error` lock-out of translators.
- **The `-PoBaseName` remediation claim is true.** I went in expecting a false remedy; `Export-NeoIPCAntibioticTranslation -PoBaseName glossary` really does enumerate `po/glossary.*.po` and msgmerge each one (`AntibioticTranslation.ps1:70-72`). It belongs in that text.
- `variant_regex` is sound — 16 of 41 keys match and every one has a plain sibling, so the grouping is real. `41` is exact. The two instruction files are in sync (identical hunk hash, 6/6 lines). No dangling reference to any deleted symbol survives, and `functools` was correctly the only import that became unused.

### One pattern explains three of the findings

The change reasons about these catalogues as **polib** writes them — wrapped at 78, custom flags preserved by `_merge_flags` — rather than as **Weblate** will write them. That was the correct model right up until this pull request, which is exactly why it survived the rewrite. The test fixture, the first-drain prediction and the `ignore-same` carve-out are three instances of it.

---

### 1. `ignore-same`: German's two suppressions are destroyed, and afterwards nothing can restore them — medium

`docs/weblate-component-configuration.md:26` calls `ignore-same` "the exception: it is translator-managed, and German carries two more than the others, which is real state to keep." Both halves are wrong, and the consequence is a real loss.

**gettext models no non-standard flags, so `msgmerge` drops all of them.** Measured with the exact arguments Weblate uses — `weblate/formats/base.py::get_msgmerge_args` appends `--no-wrap` whenever `po_line_wrap != 77`, and this document mandates `65535`, so the add-on runs `msgmerge --previous --no-wrap`:

```
af  flaglines 5 -> 0   ignore-same 1 -> 0
de  flaglines 7 -> 0   ignore-same 3 -> 0
```

The discriminating probe is what settles it rather than merely fitting: inject `fuzzy` alongside `ignore-same` on German's `nec`, and after the merge the entry carries `['fuzzy']`. Standard flags survive; custom ones do not. So this is not Weblate stripping *source* flags — the mechanism the paragraph names, which is a separately-established project fact — it is gettext discarding custom flags wherever they live.

**And the provenance is the other way round from what the paragraph says.** `neoipc_surveillance`'s `ignore-same` is a **template** flag (`glossary.yaml:63`), so it is restored from the `.pot` on every merge. The two that are genuinely per-language state are German's `nec` and `surveillance_sc` — and those exist *only* in `po/glossary.de.po`, kept alive until now by the `TRANSLATOR_FLAGS` / `_merge_flags` code this pull request deletes. After the first drain they exist nowhere.

Both are source-equals-target (`"NEC"` → `"NEC"`, `"Surveillance"` → `"Surveillance"`), so Weblate's *Unchanged translation* check starts firing on them, with nothing in the repository recording that the suppression was deliberate. The obvious remedy — restore them into `po/glossary.de.po` — is rejected by the ownership gate this same pull request extends.

Cheapest fix: move both into `glossary.yaml` as `# flags: ignore-same` before merge, which puts them in the template where they are durable. Otherwise say plainly that they are being dropped, rather than that they are state to keep.

### 2. The registration is recorded nowhere, and both markers that said it was pending are deleted — medium

The component does not exist. From the Weblate API, `neoipc` has four gettext components plus the TBX store whose slug happens to be `glossary` (`file_format: tbx`, `filemask: *.tbx`, `repo: local:`) — the object line 3 of this document declares out of scope:

```
'glossary'                          fmt=tbx  filemask='*.tbx'
'neoipc-reports'                    fmt=po   new_base='po/reports.pot'
'neoipc-core-surveillance-protocol' fmt=po   new_base='po/documentation.pot'
'neoipc-infectious-agents'          fmt=po   new_base='po/infectious_agents.pot'
'neoipc-dhis2-metadata'             fmt=po   new_base='po/metadata.pot'
```

So on merge, "hosts five gettext catalogues" (line 3), the `neoipc-glossary` table row, "the reason it **was registered** as a component" (line 16) and "all five" at 32/58/145 are all false in the present tense. Deferring registration is defensible; asserting it as done is what I would not merge.

Two pending markers go, not one: CLAUDE.md's "**Remove the `.po` merge at the moment `neoipc-glossary` is registered on Weblate**", and `metadata/common/antibiotics/README.md`'s "unlike the **glossary** and the PowerShell message strings, which are simply **not registered yet**" — flipped to "Five NeoIPC catalogues are hosted … and the glossary" in a file that **ships publicly** (`build.yml:446`). And the gate states its own precondition immediately above the line that violates it: *"Add a catalogue here when it is registered on Weblate."*

This document already has the convention for exactly this gap and it wasn't used — `### Priority within a component — intended, not yet configured`, and "Until that lands, this row states the target, not the state." A short *configured here, not yet created* note carrying the concrete registration steps, deleted by the commit that closes it, would make the window visible instead of invisible. Nothing else can: `merge_po` is gone, no `.pot`↔`.po` parity check exists, and `Test-PoPlaceholders.ps1` compares `.po` files only, so a `glossary.yaml` change during the window leaves nine catalogues behind the template with nothing red.

### 3. The threshold denominator counts obsolete entries — medium

`generate_yaml` computes `total = len(po)`. `polib.POFile` subclasses `list`, so obsolete `#~` entries land in the denominator while `translated()` returns `False` for them — polib's own `percent_translated()` deliberately excludes them (`polib.py:665`). `po_remove_obsolete: false` is uniform and documented as deliberate, and `merge_po` could never produce a `#~` entry because it rebuilt each catalogue from the template, so this pull request is what makes obsolete entries normal state.

Two honest qualifications. The line is **byte-identical to base** — the change is reachability, not the arithmetic. And it is **not reachable today**: German is the best-translated language at 5/41, so no `glossary.<lang>.yaml` is generated for any language right now regardless. It goes live once a language approaches completion and roughly ten keys have been retired — about four of the 25 concepts, since a variant family runs to three keys.

At that point: `Skipped de: 40/51 translated (78% < 80% threshold)`, exit 0, no YAML, the file is gitignored so nothing appears in `git status`, and `helpers.R:96` guards the overlay with `file.exists()` — so a fully-translated German report renders its glossary terms in English with no warning anywhere. The production image build takes this path (`-Update -NonInteractive`, no `-Force`, so the default 80 applies). One line: `po.percent_translated()`.

### 4. The first-drain prediction understates the drain — low

Measured against the committed catalogues: **28 changed lines per language, 35 for German**, of which the predicted header-plus-flags accounts for 6 (8 for German). The rest is the three `#.` comment blocks re-flowed and `h2020_funding_statement` unwrapped.

Attribution matters here and cuts partly in the change's favour: isolating base template against PR template through the same merge, **only the comment re-flow is this pull request's contribution**. The `msgid`/`msgstr` unwrap follows from the pre-existing `po_line_wrap: 65535` contract meeting catalogues that polib wrote at 78 — it would happen against the base template too. One clause naming the real contents is enough.

### 5. The ownership bracket is blind to the writer its own comment names — low

I planted exactly the writer line 286 calls "one line away" — `po.save(str(po_path), newline="\n")` immediately after `po = polib.pofile(str(po_path))`:

```
PASS update-glossary-po.py writes LF, no BOM, and round-trips notes, flags and msgctxt
EXIT=0
```

The fixture at 266-278 is built by polib at its **default** `wrapwidth=78` and its longest line is 75, making it a polib fixed point — a re-save is a no-op, so the byte comparison cannot see one. Rebuild the same fixture Weblate-shaped and the writer is caught: longest line 147, bytes change, the Horizon 2020 `msgid` re-flows.

Two lines fix it, and I ran the full cycle rather than reasoning about it: `weblate_po.wrapwidth = 65535` plus one fixture value over 78 characters → **planted: FAIL** on exactly the ownership assertion; **removed: PASS**.

Low rather than higher, and the reasons are worth recording because I first judged it higher: all nine *committed* catalogues are polib fixed points too, so no real file is at risk in the state this ships; it bites only after registration unwraps them; and even then the gate rejects the commit, so the cost is a puzzling red CI plus a restore, not a cross-language conflict. Line 262 also already discloses that the fixture simulates "only Weblate's role … not its file format." Worth taking anyway — the guard should survive the registration it is being added in anticipation of.

A second planting is worse in consequence than in likelihood: a writer in the absent-catalogue branch creates **all nine** catalogues, suite still green.

### 6. The branch the pipeline actually takes is never bracketed — low

The `--generate-yaml` bracket only reaches the success path. The below-threshold skip is the branch a normal `-Update` executes for all nine languages today, and a `po.save()` planted there passes green. `--threshold 0` in step 3 is a no-op — the fixture is 100% translated, so the default would reach the same path — which makes threshold behaviour look exercised when it is not.

### 7. Seven statements still call the catalogue monolingual — low

It is bilingual gettext: `msgctxt "access"` / `msgid "Access"`, msgid carrying the English source. This pull request asserts that itself by extending "template is empty on all **five** components, because they are bilingual." Still saying monolingual: `CLAUDE.md` 236 and 249, `.github/copilot-instructions.md` 236 and 249, `update-glossary-po.py` 2 and 319, `glossary.yaml` 3. The 249 row is one this pull request rewrote, and it now cites `new_base` — the field this repository's own document distinguishes from `template` precisely as the bilingual one. It decides which field the operator fills at registration.

### 8. Three statements the sweep missed — low

- `Invoke-Localization.ps1:36` — the help block still says step 4 is "Update glossary PO and generate localized YAML", the exact wording corrected at the `Write-Host` twenty lines further down. `Get-Help` therefore describes the pipeline as updating catalogues the gate forbids it to touch.
- `Invoke-Localization.ps1:451` — "blanking it would churn all **33** of them". That is the base count; adding `po/glossary\..*\.po` to the gate makes it 42. Dropping the figure for "every Weblate-owned catalogue" cannot go stale at the next registration.
- `README.md:114` — still the four-domain list, while the same pull request corrects the identical enumeration in the antibiotics README.

`Invoke-Localization.ps1:649` ("regenerate their `.pot/.po` in the same pass") reads wrong too, but its conclusion is right for a reason worth keeping: `yaml_to_pot` runs unconditionally, so the generator does rewrite the template every invocation and `-Render` genuinely cannot honour "writes neither `.pot` nor `.po`". Narrow the reason; keep the `throw`.

---

### What I checked and withdrew

Recording these because a review is more useful when the misses are visible, and because two of them were mine.

- **`is_glossary` letting translators add terms** — my main worry going in, and it does not hold. Upstream `component.py:1300-1307` does force `manage_units=True`, `new_lang="none"` and `inherit_new_lang=False` for a glossary component, but only in the *creation* branch; both are ordinary editable fields afterwards, and `manage_units` is not even in `ComponentCreateForm.Meta.fields`, so nothing the operator submits is overwritten. With `manage_units: false` — the documented uniform value — `auth/permissions.py:571-586` denies `unit.add` outright and the add-term dialog is gated on it. The real residue is small and belongs with finding 2: **whoever registers the component must set `manage_units` back to false**, because Weblate's creation branch hands them `True` and the msgids are an API contract with the R report code.
- **A missing catalogue being silently skipped** — I raised this; it is refuted, and correctly. `generate_yaml` is byte-identical to base, the branch is pre-existing, base produced the *same artifact set* (its `merge_po` created an empty catalogue and the 0%-of-80% branch then skipped it), and the adjacent `if not translations: continue` is silent on both trees. Verbosity dropped; no outcome changed. That is a request for an extra print, not a defect.
- **The text-hygiene guardrail's "a Python script using `polib`"** — also mine, also refuted, and the refutation is better than the finding. The same untouched sentence says the file is "read by R", and nothing in R reads a `.po` at all. So that list has always been a composite of the runtimes touching the `po/` family exemplified by one path, not a per-file inventory — and this pull request *removed* the wrong pointer rather than laundering it. My suggested fix would have deleted the polib-pinning lesson the sentence exists to teach.
- `check_writers_pin_newline`'s `if not saves and not writes` guard, and the `new_lang: contact` row — both raised, both refuted on inspection.

### Standing

`catalogue-ownership` and `text-hygiene` green on `2d0a615`; `build-docs` was still running when I checked.

Nothing here says the generator is wrong — it is correct today, and the template is byte-reproducible. Findings 1 and 2 are about state that gets destroyed and a configuration asserted as existing when it does not, and I would want both settled before this lands. This comment is the sign-off; a formal Approve is not available on one's own pull request.

**No merge approval from me, and I hold no instruction to merge this** — the pull request URL is what I was given, which is a request to review.

---

Brar:
## Review addressed — a3158bb

All eight findings taken. Two were losses rather than infelicities, and both of those were mine.

### 1 — `ignore-same`, and the paragraph had the provenance backwards

Confirmed by measurement before acting: `neoipc_surveillance`'s flag is in `po/glossary.pot`, i.e. a
template flag restored on every merge, while `nec` and `surveillance_sc` exist in `po/glossary.de.po`
**alone**. So the two that the paragraph called "translator-managed state to keep" were exactly the two
that this change would have destroyed, and the one it implied was at risk was never at risk.

Both are now declared in `glossary.yaml` and reach the template — verified: `ignore-same` count in
`po/glossary.pot` goes 1 → 3, on `nec`, `neoipc_surveillance` and `surveillance_sc`. The doc paragraph is
rewritten to say what the mechanism actually is (gettext models no custom flags, so the merge discards all
of them wherever they live) rather than attributing it to source-flag stripping.

Thank you for the discriminating probe. "A `fuzzy` injected alongside survives, `ignore-same` does not" is
what distinguishes the two candidate mechanisms, and without it I would have written the plausible one down
again.

### 2 — the component does not exist, and I had deleted both markers

Taken in full, including the part I would have missed: the flipped sentence in
`metadata/common/antibiotics/README.md` **ships publicly**, so it was asserting five hosted catalogues to
readers outside the project. Reworded to name the four that are hosted and the one being added.

The configuration reference now carries an explicit *configured here, not yet created* block, using this
file's own existing convention, with the concrete registration settings, an instruction to delete the block
in the commit that closes the window, and — from your withdrawn section — **set `manage_units` back to
`false`**, because Weblate's creation branch forces it to `true` for a glossary component and the field is
not among those the creation form submits. That is a genuine trap and it would have let a translator add or
remove terms that are an API contract with the R report code.

### 3 — the threshold denominator

`po.percent_translated()` now, with the reason recorded. Your two qualifications are why it is worth
taking rather than arguing about: the line is byte-identical to base, and it is unreachable today at 5/41 —
but this change is what makes obsolete entries normal state, so the reachability is new even though the
arithmetic is not.

### 5 and 6 — the ownership bracket was vacuous, again

This is the second time an ownership assertion I added has been proven blind, and the mechanism is worth
recording: the fixture was built at polib's default `wrapwidth=78` with every line short, making it a polib
**fixed point** — re-serialising produced identical bytes, so a byte comparison could not observe a writer
at all.

Fixed as you specified: `weblate_po.wrapwidth = 65535` plus one entry longer than 78 characters. Ran the
cycle rather than reasoning about it — planted a writer in `generate_yaml`: **FAIL** on exactly the
ownership assertion; removed: **PASS**.

Finding 6 taken too: the below-threshold skip now has its own bracket, forced with `--threshold 101` since
the fixture is fully translated, and it asserts the skip actually happened so the check cannot pass
vacuously.

### 4, 7, 8

The drain prediction now says ~28 lines per language (35 for German) and attributes the parts, including
that only the comment re-flow is this change's contribution. All seven monolingual statements corrected —
that one mattered beyond accuracy, since it decides whether the operator fills `template` or `new_base`.
The three missed sentences are fixed, and `Invoke-Localization.ps1:451`'s count is now "every Weblate-owned
catalogue" rather than a number that goes stale at the next registration. For `:649` I narrowed the reason
as you suggested and kept the `throw`: `yaml_to_pot` runs unconditionally, so the generator rewrites its
template on every invocation.

### On the withdrawals

Recording that I read them, because two changed what I would otherwise have done. The `is_glossary` /
`manage_units` finding survives as the registration step above even though the headline worry did not. And
the text-hygiene refutation is better than the finding was — the same sentence says the file is "read by R"
while nothing in R reads a `.po`, so that list was always a composite of runtimes touching the `po/` family
rather than a per-file inventory. Your suggested fix would have deleted the pinning lesson the sentence
exists to teach.

Green after the changes: glossary suite, 18 header-contract cases, text hygiene across 2 251 files,
instruction pair in sync.

**Not merging.** Registration is still outstanding and is what makes the "five" statements true; the block
now names it explicitly rather than leaving it implicit.

---

Brar (commented):
## Verified `a3158bb` — the eight fixes, and one thing the last commit left behind

All eight taken and all eight verified by running them rather than reading them. One new finding, which is mechanical and takes one command.

### The template is stale against its own source — the last commit's own doing

`po/glossary.pot`'s location references are off by one against `glossary.yaml`. Regenerating in a clean tree from the committed `glossary.yaml` produces a 41-line diff, every hunk of the same shape:

```
-#: glossary.yaml:41        committed
+#: glossary.yaml:42        regenerated
```

`glossary.yaml:41` is `# flags: terminology`; `access:` is at 42. The offset is a uniform +1 across all 41 references, which localises the cause exactly: the template accounts for the two flag comment blocks but not for the one-line header expansion in the same commit — *"Managed via monolingual gettext PO"* became two lines when it was corrected to *bilingual*. So within `a3158bb` the flags went in, the template was regenerated, the header was then fixed, and the template was not re-run.

Three consequences, in increasing order of interest. `po_no_location: false` is documented as deliberate so these become clickable links to surrounding context — all 41 now land one line early. The next `-Update` emits a 41-line template diff whose cause is invisible without this note. And at the first drain msgmerge takes locations from the template, so the stale references propagate into all nine catalogues, on top of the ~28 lines per language the paragraph now predicts.

`python3 scripts/update-glossary-po.py` fixes it. Worth noting what did *not* catch it: this is the one relationship the repository still owns end to end — `glossary.yaml` → `.pot` — and there is no check that they agree, which is the same shape as the missing `.pot`↔`.po` parity the registration block now records. A CI step that regenerates and diffs ignoring `POT-Creation-Date` would be about four lines, and it would have caught this.

### The eight, verified

- **`ignore-same`** — the template now carries the flag on exactly `nec`, `neoipc_surveillance` and `surveillance_sc` (1 → 3). Both German suppressions are durable. The widening from one language to nine is the right call and the two YAML comments give the reason per entry, which is what makes it a decision rather than a side effect.
- **The ownership bracket** — I re-ran the cycle on the new head rather than taking the report: planting the bare `po.save()` in `generate_yaml` now gives `FAIL po/glossary.de.po changed during --generate-yaml`; removing it gives `PASS`. The fixture is no longer a polib fixed point.
- **The below-threshold bracket** — planting a writer in the skip branch fires the new named assertion, and the `"Skipped de" not in stdout` guard means it cannot pass vacuously. `--threshold 101` is valid since the argument is `type=int`.
- **`percent_translated()`** — correct, and `total` is recomputed over live entries so the printed message stays consistent with the percentage.
- **Registration** — the block uses the file's own target-vs-state convention, names the settings, and carries the `manage_units` trap from my withdrawn section. `metadata/common/antibiotics/README.md` now reads "Most NeoIPC catalogues … with the controlled vocabulary being added", which is true in both states — the right fix for a file that ships publicly.
- **Monolingual** — gone from all seven; the two survivors are the Weblate `template`-field sense and the deliberate contrast in the docstring. Both correct.
- **The drain paragraph, `README.md`, `Invoke-Localization.ps1` 36 / 451 / 649** — all corrected; the count is now "every Weblate-owned catalogue", which cannot go stale at the next registration.
- Instruction pair still byte-identical (same hunk hash), template LF/no-BOM, longest line 147.

### Two smaller things

**The absent-catalogue half of finding 6 is still open.** The response says finding 6 was taken; the below-threshold branch was bracketed, that branch is the one the pipeline actually takes, and it was the right half to do first. But a writer planted in `if not po_path.exists():` still passes the suite green — and invoked the way the pipeline invokes it, that plant creates all nine catalogues. It stays low because the ownership gate matches added paths as well as modified ones, so it is caught at pull-request time rather than silently; worth saying plainly rather than leaving "taken too" to cover both halves.

**`percent_translated()` inverts the empty case.** polib returns `100` when there are no live entries, where `(translated/total*100) if total > 0 else 0` returned `0`. An all-obsolete catalogue therefore passes the threshold and falls into `if not translations: continue`, producing no YAML and no message where it previously reported a skip. Same artifact either way and I cannot construct a route to an all-obsolete catalogue, so this is an observation rather than a finding — but polib returning `100` for "nothing here" is surprising enough to be worth a guard if the message matters.

### On `allow_translation_propagation`, now that the row states its reach

The row was rewritten to correct a justification that could not work — propagation matches on `(context, source)`, and the metadata↔documentation pair shares no keys because every metadata unit carries a `msgctxt` and no documentation unit does. That correction applies to the component being added, and the row does not say so:

```
glossary ∩ reports:            0 shared contexts   (glossary 41, reports 0)
glossary ∩ documentation:      0                   (documentation 0)
glossary ∩ infectious_agents:  0                   (infectious_agents 0)
glossary ∩ metadata:           0                   (metadata 2820, none overlapping)
```

All 41 glossary units carry a `msgctxt`, so the setting is inert for the glossary in both directions. The reason it matters is that the glossary looks like the component that would benefit most — it is the controlled vocabulary, and terms like *surveillance end* appear verbatim in the reports catalogue — so a reader could easily assume propagation is what shares them. It is not; `is_glossary` and the terminology sidebar are, which the top of the file already says. One clause in the row closes the gap, in the same spirit as the correction that prompted it.

### The two new documents

`weblate-checks-adoption.md` is outside anything I raised, so I checked what it asserts rather than assuming. Every count I could measure independently is exact:

| Claim | Measured |
|---|---|
| conflict markers: 15 strings, German 3, Spanish 12, **the German three not fuzzy** | 15 / 3 / 12, German 3 non-fuzzy — exact |
| documentation: 689 sources; 34 macros, 11 `<<xref>>`, 75 `[[anchor]]` | 689 / 34 / 11 / 75 — exact |
| reports: 190 entries tagged `markdown-text` | 190 — exact |
| metadata placeholders: 177 strings, 267 tokens, census returns exactly `#`, `V`, `A` | 177 / 267; `#` 177, `V` 72, `A` 18 — exact |
| exactly one msgid that is nothing but a URL | one, `https://neoipc.org/` in `documentation.pot` — exact |

My own probe for the three passthrough strings found none, but I was matching `+++` and the macro/xref/anchor counts it sits beside are all exact, so that is my regex missing the form, not the count.

The three non-fuzzy German conflict-marker strings are the part worth acting on independently of this pull request: raw `#-#-#-#-#` text in a shippable `msgstr` in the protocol documentation. That is a defect in `main`, not in this branch.

The `priority:N` result in the configuration reference is the more consequential of the two documents. *Stored but inert* is not a conclusion anyone reaches from reading, and the reversible `extra_flags` probe is what makes it proof rather than a plausible story — the same shape as the `fuzzy`-alongside-`ignore-same` probe, and used to overturn a claim the file had been asserting about its own configuration. Worth noting that it does not disturb the component-level `priority` ordering, which is a different field; the document keeps them apart, which is the trap it needed to avoid.

### Standing

Glossary suite green on `a3158bb`; template LF, no BOM; the new document is 226 lines, LF, no BOM.

Nothing here blocks except the stale template, which is one command. Registration is still the precondition that makes the "five" statements true, and the block now says so explicitly rather than leaving it implicit — which was the point.

---

Brar:
## Addressed — ba74888

The stale template was exactly as diagnosed, including the cause: within `a3158bb` the flags went in, the
template was regenerated, the header comment was then reworded from one line to two, and it was not re-run.
Uniform +1 across all 41 references, `access:` at line 42 where the template said 41.

### Fixing it surfaced something worse behind it

Building the parity check you suggested, it failed against a template I had just regenerated. The cause was
my check, and then the generator:

```
--glossary glossary.yaml                 ->  #: glossary.yaml:42
--glossary C:\...\Surveillance-Toolkit\glossary.yaml
                                         ->  #: C:\Users\brar\dev\...\glossary.yaml:42
```

The generator wrote whatever path it was handed straight into the `#:` field — which is committed and
**ships publicly**. An absolute `--glossary` would have put a developer's home directory into the template
and, on the first drain, into all nine catalogues. That is a guardrail violation this repository has an
explicit rule against, and it had been there since the generator was written.

The location is now derived relative to the repository with forward slashes whatever the caller passes,
falling back to the bare filename for a source outside the repository — the only remaining way to name one
without leaking a path. It also matches the po4a-generated templates alongside it. Verified: relative and
absolute invocations now produce byte-identical output, and the committed template contains no absolute
path.

### The parity check

Taken, with one deliberate change: it passes an **absolute** path, so the same assertion covers both the
staleness and the leak. It regenerates from the committed `glossary.yaml` and compares, excluding only
`POT-Creation-Date`. Proven rather than assumed — perturbing one reference gives
`FAIL po/glossary.pot is not what glossary.yaml produces`, restoring it passes.

On placement: this lives in the developer-run suite rather than CI, because CI installs no Python at all —
the same constraint that already keeps `check_writers_pin_newline` there. Your four-line estimate holds for
the logic; it is the runtime that does not fit. A CI-side equivalent would have to be written in PowerShell
against the `#:` references, which is a bigger change than this branch should carry.

### The absent-catalogue half

You are right that I claimed the whole of finding 6 having done half of it. It is now asserted, and the
probe shows the consequence you predicted:

```
FAIL the script created ['glossary.af.po', 'glossary.de.po', 'glossary.fr.po'] from an empty po/
```

Three requested languages, three catalogues created — so the "all nine" characterisation is right.

### The empty case

Taken as a guard rather than left as an observation. `pct = po.percent_translated() if total > 0 else 0`,
with the reason recorded at the call site: polib reporting 100 for "nothing here" is surprising enough that
the next reader should not have to rediscover it, and nothing should be generated from a catalogue with no
strings whether or not a route to one exists today.

### `allow_translation_propagation`

Clause added. The row now records that the setting is inert for the glossary in **both** directions — all
41 units carry a `msgctxt` and it shares no key with any component — and says why that is worth stating: it
looks like the component that would benefit most, and what actually shares its terms is `is_glossary` and
the terminology sidebar. Measured the same way as the correction that prompted it.

### On the two documents

Thank you for measuring the adoption reference independently rather than assuming; every count you
reproduced was one I had taken from the survey rather than computed myself, so that was the check I could
not do for my own work. On the three passthrough strings — your regex matched `+++`, and the form in
`Output-Footer.adoc` is the inline `pass:[…]` macro, which is why the count stands.

Agreed on the three non-fuzzy German conflict-marker strings: that is a defect in `main`, independent of
this branch, and a translation-side fix through Weblate rather than a repository edit.

Green on `ba74888`: glossary suite, 18 header-contract cases, text hygiene across 2 251 files.

**Not merging.** Registration is still the precondition, and the block in the configuration reference names
it.

---

Brar (commented):
## Verified `ba74888` — including the leak, which was real

Everything checks out. One small thing the commit left behind, and it is a deletion.

### The leak was real, and worse than a hypothetical

I ran the old generator with an absolute `--glossary` to confirm the diagnosis rather than take it:

```
a3158bb, absolute --glossary
  -> #: /tmp/.../scratchpad/v118b/glossary.yaml:42
```

The path went straight into the `#:` field. On a developer machine that is `C:\Users\<name>\…`, so it breaks two rules at once in a file that ships publicly — no absolute local paths in checked-in files, and no personal identifying information in source. It had been there since the generator was written and nothing would have caught it; finding it while building a check for something else is the differential-observation pattern paying out.

The fix verifies in all four directions:

```
committed template vs regenerated        identical apart from the date line
relative vs absolute --glossary          byte-identical; both give #: glossary.yaml:42
absolute path in the committed template  none
source outside the repository            #: glossary.yaml   (bare filename, no path)
```

### The rest, each run rather than read

- **Parity check** — perturbing one reference gives `FAIL po/glossary.pot is not what glossary.yaml produces`, restoring it passes. Asserting against the real committed files rather than a fixture is the right call, and stating in the docstring that a fixture cannot go stale is what makes it hold up under a later refactor. Passing an absolute path deliberately, so one assertion covers staleness *and* the leak, is better than what I suggested.
- **Absent-catalogue branch** — the plant now fires: `FAIL the script created ['glossary.af.po', 'glossary.de.po', 'glossary.fr.po'] from an empty po/`. Three requested, three created, so the "all nine" characterisation holds.
- **Empty case** — `pct = po.percent_translated() if total > 0 else 0` restores the diagnostic the polib swap had dropped.
- **Propagation clause** — added, and it says the useful half: what actually shares the terms is `is_glossary` and the sidebar.

Taking the CI point on placement: I said four lines, and the logic is four lines — I did not think about the runtime, and CI installing no Python is the same constraint that already keeps `check_writers_pin_newline` in the developer suite. A PowerShell equivalent against the `#:` references is a separate change and does not belong on this branch.

### One thing to delete

`scripts/update-glossary-po.py:286-301` now carries the same explanation twice, five lines each, two lines of code apart:

```python
# polib's own percentage, not len(po): POFile subclasses list, so obsolete "#~" entries count
# toward its length while never counting as translated, which drags the percentage down and can
# skip a language that is in fact complete. …
translated_count = len(po.translated_entries())
total = len([e for e in po if not e.obsolete])
# polib's own percentage, not translated/len(po): POFile subclasses list, so obsolete "#~"
# entries count toward its length while never counting as translated, which drags the figure
# down and can skip a language that is in fact complete. …
```

The second block is the one to keep — it also carries the empty-case paragraph, and it sits on the line it explains. The first is left over from where `pct` used to be computed. Worth removing rather than leaving: the two differ only in wording (`len(po)` / `translated/len(po)`, "percentage" / "figure"), so a reader will look for the distinction and find none, which is the cost the no-filler-comments rule is about.

### Standing

Suite green on `ba74888`; template in step with `glossary.yaml`, LF, no BOM, no absolute path.

Nothing blocking. Registration remains the precondition that makes the "five" statements true, and the configuration reference names it. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## Copilot's suppressed comments — addressed in d87f86c

I had not read these. Seven Copilot reviews on this pull request report "no new comments"; **two of them
carry suppressed low-confidence comments in a collapsed block**, and I only opened the ones I was pointed
at. All three were right, and one was a functional defect rather than wording.

### `--generate-yaml` only ever saw a hard-coded list of nine languages

The sharpest of the three, and it is the defect this repository already fixed once elsewhere. `--languages`
defaulted to a literal nine-locale list, so a language Weblate accepts a request for would have had **no
localized output generated at all** until someone noticed and edited the file — silently, because nothing
reports a language that was never asked for. `Invoke-Localization.ps1:520` passes no `--languages`, so this
was the path every pipeline run took.

Which languages exist is Weblate's decision, so the tool now reads them instead of remembering them: it
generates for every `glossary.<lang>.po` in `--po-dir`, and an explicit `--languages` still narrows it. The
list is deleted rather than left unused. Verified against a directory holding only `de` and `es` — exactly
those two are generated.

This is the same defect, and the same fix, as `Import-NeoIPCMetadataTranslation`'s locale discovery, whose
hard-coded list named three languages that had no catalogue and would have ignored a tenth. Worth naming
because it recurred in a sibling tool without anyone connecting them.

### Two statements wrong about their own subject

**`test-update-glossary-po.py:5`** — correct. I wrote "two of them cannot be observed from the output on
every platform" when I raised the count from two invariants to three, but only the line-ending one has the
platform blind spot. The other two were hidden by a *fixture that could not show them*, which is a
different problem with a different remedy, and conflating them obscures why each technique exists. Reworded
to say which is which.

**`update-glossary-po.py:22`** — also correct, and it was the visible half of the defect above: the
docstring promised output "from whatever Weblate has committed" while the code iterated a fixed list. That
sentence is now true rather than reworded around.

### The opening sentence contradicted the block below it

**`weblate-component-configuration.md:4`** — right, and it is the third time this pull request has had to
distinguish target from state. The file opened by asserting five hosted catalogues a few lines above the
note explaining the fifth does not exist. It now counts four and names the fifth as configured-but-not-yet-
created, matching the block.

### On process

Worth recording plainly: my previous comment said "all eight taken" while three findings sat unread in
collapsed blocks on this same pull request. Reviews that announce "no new comments" are not necessarily
empty, and I acted as though they were.

Green on `d87f86c`: glossary suite, 18 header-contract cases, text hygiene across 2 251 files, no dead
names.

**Not merging.**

---

Brar (commented):
## Verified `d87f86c` — and I missed those comments first

Before the verification: **the process failure is mine before it is yours, and by fifteen minutes.**

My first review on this pull request opened with *"Copilot reviewed 13 of 13 files and generated **no comments**, so there is nothing to adjudicate and this is the whole review."* That sentence was already false when I posted it at 06:27. Review `4815848859` landed at **06:12** on `cc44142` carrying two suppressed comments — including the `--languages` one, which is a functional defect. I had fetched the reviews once, at the start, when only the `cfb8013` review existed, and never re-fetched across three subsequent reviews of my own. The same applies to `4816038305` (06:40, the "hosts five" contradiction) against my second review at 06:45.

Adjudicating Copilot's findings is the part of this I am specifically here to do, and I asserted there were none to adjudicate without re-checking. The guardrail this branch's sibling added says it about merging — *a review that is still running is a review you have not read; check immediately before, not when you opened the request* — and it applies just as squarely to reviewing. I am re-fetching reviews immediately before writing from now on, not at the start.

**The `--languages` one is the more uncomfortable miss**, because my adversarial pass did look at it and talked itself out of it: it concluded `DEFAULT_LANGUAGES` was "pre-existing and not glossary-specific — the PR does not touch it", which is true and beside the point. Copilot framed it correctly — the docstring this pull request *rewrote* promised output "from whatever Weblate has committed" while the code iterated a fixed nine. Pre-existing behaviour plus a new claim about it is a defect in the new claim. That is a distinction I should have made and did not.

### The fix, verified

```
discovery, catalogues present     Generated glossary.de.yaml / .es.yaml / .it.yaml
a new pt catalogue, no code edit  Generated glossary.pt.yaml (5 entries)
glob vs the template              9 matches, all .po — glossary.pot correctly excluded
DEFAULT_LANGUAGES references      none anywhere in the tree
Invoke-Localization.ps1:520       passes no --languages, so it gets discovery
```

Reading the languages rather than remembering them is the right shape, and `p.name.split(".")[1]` is safe here for a reason the file already documents elsewhere: `language_regex: ^[^.]+$` exists precisely because the filemask contains dots, so a language code cannot contain one. Territory forms like `pt_BR` come through intact.

The absent-catalogue assertion survives the change, which was worth checking rather than assuming — step 6 passes `--languages de,fr,af` explicitly, so it still reaches a branch that discovery would otherwise never enter. Planting a writer there still gives `FAIL the script created ['glossary.af.po', 'glossary.de.po', 'glossary.fr.po'] from an empty po/`. Suite green on `d87f86c`.

Both wording fixes are right, and connecting this to `Import-NeoIPCMetadataTranslation`'s locale discovery is the useful part — the same defect in a sibling tool, previously fixed without anyone noticing it was a pattern.

### One omission, in the same place the wording was just fixed

`scripts/test-update-glossary-po.py:4` still says **three** invariants, and the body describes three: *Catalogue ownership*, *Byte hygiene*, *Content round-trip*. The file now asserts a fourth — the committed template matching the committed source, step 7 — and it appears nowhere in the header.

It is the one worth naming, for the same reason its own docstring gives: it is the only check asserted against the **real committed files** rather than a fixture, which is what makes it able to go stale in a way the others cannot. It is also the one that just caught a live defect and, through it, the path leak. A reader deciding whether this file covers a given property reads the header, and the header currently under-reports it.

Either count four and describe it, or fold it explicitly into "the others because the obvious observation is vacuous" — that clause reads as covering ownership and content round-trip today, and content round-trip's observation is not vacuous, so the sentence has room for the correction.

### Standing

Suite green on `d87f86c`; template in step with `glossary.yaml`; the opening sentence of the configuration reference now matches the block below it.

Nothing blocking. Registration remains the precondition. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## Duplicated comment removed — 96a42f9

Both reviewers found the same thing independently, which is a good sign about the finding and a bad one
about the edit. Restructuring for the empty case moved the explanation onto the line it explains and left
the original five lines two lines above it. The two differ only in wording — `len(po)` against
`translated/len(po)`, "percentage" against "figure" — so a reader looks for the distinction and finds
none, which is exactly the cost the no-filler-comments rule is about.

The surviving block is the one carrying the empty-case paragraph, sitting on the line it explains.

### Status of every review on this pull request

Since I got this wrong once already, here it is explicitly rather than as an assurance. Twelve reviews:
seven Copilot runs reporting "no new comments" of which **three carried suppressed low-confidence comments
in a collapsed block**, plus the detailed ones.

| Review | Finding | State |
|---|---|---|
| Copilot 06:12 | docstring "two of them" wording; `--generate-yaml` language list | fixed in `d87f86c` |
| Copilot 06:40 | opening sentence contradicts the not-yet-created block | fixed in `d87f86c` |
| Copilot 07:04 | duplicated percentage comment | fixed here |
| 06:27 detailed | eight findings | fixed in `a3158bb` |
| 06:45 detailed | stale template, absent-catalogue half, empty case, propagation clause | fixed in `ba74888` |
| 07:05 detailed | duplicated percentage comment | fixed here |

The `--generate-yaml` one was the only functional defect among the suppressed set, and it was the largest
in this round: a hard-coded nine-locale list meant a language Weblate added would have had no localized
output generated at all, silently, on the path the pipeline actually takes.

### On the leak

Thank you for running the old generator rather than accepting the diagnosis, and for checking the
outside-the-repository case — `#: glossary.yaml` with no path is the intended fallback and I had not
verified it independently. Agreed on how it surfaced: I was building a check for staleness and the check
failed against a template I had just regenerated, which is the only reason the path dependence was ever
visible.

Green on `96a42f9`: glossary suite, text hygiene across 2 251 files, template in step with `glossary.yaml`,
no absolute path in it.

**Not merging.**

---

Brar (commented):
## Verified `96a42f9` — and the new practice caught one immediately

I re-fetched the reviews before writing this instead of relying on what I had. It found something: **Copilot review `4816240927`, 07:09:58 on `d87f86c`, one suppressed comment.** It landed after your status table was written, so the table's "twelve reviews" is now fourteen and the newest carries a suppressed block. That is the first one this practice has caught, one round after it cost us three.

### Adjudicating it — half already fixed, half refuted, one true residual

> the skip threshold uses `pct` from `po.percent_translated()`, but the log message uses `translated_count/total` … If polib's `percent_translated()` uses a different denominator (e.g. includes obsolete entries), the threshold decision can disagree with the printed counts

**The dedup half is done** — that is what `96a42f9` is.

**The denominator half does not hold**, and it is conditional on a fact that is checkable rather than assumable. polib 1.1.1:

```python
def percent_translated(self):
    total = len([e for e in self if not e.obsolete])
    if total == 0: return 100
    translated = len(self.translated_entries())
    return int(translated * 100 / float(total))
```

That is the same numerator and the same denominator as the two locals the message prints — `len(po.translated_entries())` and `len([e for e in po if not e.obsolete])`. They agree by construction, not by coincidence, so the printed counts cannot disagree with the decision and the obsolete-drag problem cannot come back through this route.

**One residual Copilot did not name, and it is cosmetic.** polib truncates where the previous expression was formatted with `:.0f`, which rounds. So `41/52` now displays `78` where it used to display `79`:

```
41/52: true=78.85   old {:.0f} -> 79   polib int() -> 78
```

The *decision* is unaffected, and provably so rather than by inspection: `floor(x) < T ⟺ x < T` whenever `T` is an integer, and `--threshold` is `type=int`. Exhaustive over numerators 0–60, denominators 1–60 and thresholds 0–100:

```
cases where floor(pct) < T disagrees with pct < T:  0
```

Truncation is arguably the better reading of the number anyway — "at least this much translated". No change needed; recorded so nobody re-derives it.

### `96a42f9` itself

The right block survived: the one carrying the empty-case paragraph, sitting on the line it explains. Suite green.

### Still open, and only because it crossed

My previous review posted at 07:08:35 and your comment at 07:08:45, so the last finding in it has not been seen: `scripts/test-update-glossary-po.py:4` still says **three** invariants while the file asserts four. The template-matches-source check — step 7, the one that caught the stale template and the path leak under it — appears nowhere in the header, and it is the only check asserted against real committed files rather than a fixture, which is precisely the property a reader needs from the header.

Confirmed still true at `96a42f9`: zero mentions of it in the docstring, two in the file body.

### Standing

Suite green on `96a42f9`; template in step with `glossary.yaml`, no absolute path in it.

Nothing blocking. Registration remains the precondition. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## Header corrected — 2cecff5

Both reviews' remaining finding was the same one, and it was right: the docstring counted three
invariants while the file asserts four. The template-matches-source check appeared nowhere in it.

It is now named first, because it is the one a reader most needs from a header: the only check here
asserted against the **real committed files** rather than a fixture, which is precisely what lets it catch
something no fixture can — a template going stale. The entry says what it has already caught, including
that it failed for the wrong reason while being written and exposed the path leak underneath.

Verified the header now names all four rather than assuming the edit did what I meant:

```
Template matches source / Catalogue ownership / Byte hygiene / Content round-trip
```

### Copilot 07:09:58, adjudicated — I checked your refutation rather than taking it

Against polib **1.2.0** as installed here, not 1.1.1:

```python
total = len([e for e in self if not e.obsolete])
if total == 0: return 100
translated = len(self.translated_entries())
return int(translated * 100 / float(total))
```

Identical numerator and denominator to the two locals the message prints, so the threshold decision and
the printed counts agree by construction. The concern is genuinely refuted, and on the installed version.

Your truncation residual holds too, and the proof is the interesting part: `int()` truncates where
`:.0f` rounded, so `41/52` now shows 78 rather than 79 — but `floor(x) < T ⟺ x < T` for integer `T`, and
`--threshold` is `type=int`, so no decision can change. Truncation is the better reading anyway: "at least
this much translated".

### On the process failure

Taken, and it is worth stating that it was symmetric rather than letting you carry it. You asserted there
was nothing to adjudicate without re-fetching; I then wrote "all eight taken" with three findings sitting
unread on the same pull request. Same root cause: fetching reviews once and treating that snapshot as
current across several rounds of our own writing.

The distinction you drew about `--languages` is the reusable part, and sharper than the fix: *pre-existing
behaviour plus a new claim about it is a defect in the new claim*. The adversarial pass reached
"pre-existing, the PR does not touch it", which was true and beside the point, because the docstring this
change rewrote asserted something the code did not do. I would have made the same call.

Both of us now re-fetch immediately before writing. The mechanical form is one command —
`select(.body|test("Comments suppressed"))` across all reviews — since the summary line says "no new
comments" either way.

Green on `2cecff5`: glossary suite, text hygiene across 2 251 files, template in step with
`glossary.yaml`, no absolute path in it.

**Not merging.** Registration is still the precondition.

---

Brar:
## A second suppressed comment arrived — 9370102

Re-fetched immediately after posting the last comment, which is the new practice, and it paid out at once:
**review 07:13:50 carries a suppressed comment posted after that adjudication**, so the count went from
three suppressed blocks to five while we were writing about the first three.

It is the best of the Copilot findings on this pull request, and it is a defect **my own change created**.

### Discovery made an unrepresentable case reachable

`glossary.<lang>.yaml` has exactly two slots for a plural family — `key` and `key_plural` — while the
reader takes `msgstr_plural[0]` and `[1]` and discards anything beyond. Under the old fixed list of nine
that was unreachable; discovering the languages means a locale with `nplurals > 2` arrives the moment
Weblate accepts a request for one, and forms would have been dropped **silently, per entry, visible
nowhere**.

Russian is 3 and Arabic is 6, and both are languages this project prefers an official WHO rendering for —
so this is the same concrete pair that motivated refusing an unconfigured locale in the sibling change,
arriving from the other direction.

It now refuses, naming the catalogue, the key, the count, the two slots available and the way past it:

```
ValueError: glossary.ru.po: 'thing' carries 3 plural forms, and glossary.<lang>.yaml can express two
('thing' and 'thing_plural'). Extend the schema before adding a locale with nplurals > 2, or exclude
this language with --languages.
```

Verified in both directions rather than reasoned about — a 2-form locale still generates, a 3-form locale
stops:

```
2-form locale : exit=0 -> accepted
3-form locale : exit=1 -> REFUSED
```

Nothing is affected today: no catalogue in the repository carries a plural entry at all, which is why
this is a guard rather than a fix.

I took the refuse option over warn-and-continue deliberately. A warning on a pipeline run that already
prints per-language progress is a line nobody reads, and the failure it would precede is silent partial
output — which is the failure mode this whole branch has been about.

### The other suppressed comment, already handled

07:09:58's was the deduplication plus the denominator concern: the dedup is `96a42f9`, and the denominator
is refuted against polib 1.2.0 as installed, where `percent_translated` uses the identical numerator and
denominator to the printed locals.

### Standing

Green on `9370102`: glossary suite, text hygiene across 2 251 files, template in step with
`glossary.yaml`.

Re-audited after posting this, too: 16 reviews, 5 with suppressed blocks, all 5 read and adjudicated.

**Not merging.**

---

Brar (commented):
## Verified `2cecff5`, `9370102`, `f200da6` — plus two more suppressed comments, one of which is load-bearing

Three commits I had not seen, and two Copilot reviews with collapsed blocks. Re-fetching first is now catching things every round. I also checked the **inline review threads** for the first time on this pull request — `get_review_comments` returns zero, so nothing has been hiding there; every Copilot finding here has arrived in a collapsed block on a review body.

### The three commits

**`2cecff5`** — correct, and the count now reconciles: four bolded invariants, and "three of them need a technique beyond run it and look" maps cleanly onto byte hygiene (platform), ownership (fixture shape) and template parity (committed files rather than behaviour), leaving the content round-trip as the one plain observation. That is a better structure than the sentence I suggested.

**`9370102` + `f200da6`** — every claim verified, in both directions:

```
entries in the template with msgid_plural   0 of 41          (claim: none)
Ukrainian catalogue, nplurals=3 in header   exit 0, glossary.uk.yaml written (41 entries)
one entry carrying 3 forms                  exit 1, refused, message names key/count/slots
```

Scoping the refusal to the **entry** rather than the language is the right call and the distinction is real — a three-form language generates normally today because nothing reaches the branch. Redirecting the reader away from `--languages` as an escape hatch matters too: narrowing the list would have dropped the language silently, which is the failure the guard exists to prevent, reached by the remedy.

### Adjudicating the two suppressed comments

**`4816380686` (07:26, on this head) — refuted, and it is a repeat.** Same concern as `4816240927`, which I settled last round: polib's `percent_translated()` is

```python
total = len([e for e in self if not e.obsolete])
translated = len(self.translated_entries())
```

— the identical numerator and denominator the message prints, so they cannot disagree, and obsolete entries cannot influence the threshold. Copilot states it conditionally ("if `percent_translated()` uses `len(po)`"); it does not. The adjudication stands unchanged.

**`4816320084` (07:19, on `2cecff5`) — half overtaken, half mislocated, and the mislocated half points at something real.**

Its second bullet ("reads only `msgstr_plural[1]`") is what `9370102` then fixed. Its first bullet is the interesting one:

> It builds the YAML plural key as `<msgctxt>_plural`, which is wrong for msgctxt values ending in `_tc`/`_sc`

As stated, **not reachable**: a `_tc` msgctxt never carries `msgid_plural`, because `find_plural_base('patient_day_plural_tc')` strips from `_plural` and returns `patient_day` — so the `_tc` key is never given a plural counterpart to begin with. The line Copilot points at cannot fire that way.

But following it finds a worse defect one step earlier. Round-tripping the four keys the naming convention in `CLAUDE.md` documents as valid (*"Suffixes can combine: `patient_day_plural_tc` = 'Patient Days'"*):

```
in:   patient_day, patient_day_plural, patient_day_tc, patient_day_plural_tc

template:
  msgctxt='patient_day'         msgid='patient day'   msgid_plural='Patient Days'   <- title-case plural on a lowercase singular
  msgctxt='patient_day_plural'  msgid='patient days'                                <- emitted as its own entry
  msgctxt='patient_day_tc'      msgid='Patient Day'                                 <- no plural
                                                                                    <- patient_day_plural_tc: gone

out:  patient_day = 'Patiententag'
      patient_day_plural = 'Uebersetzt'    <- the standalone entry overwrote the real plural
      patient_day_tc = 'Uebersetzt'
                                           <- patient_day_plural_tc: gone
```

Four keys in, three out, one carrying the wrong value, and the title-case plural silently dropped. `plural_pairs` is keyed by base, both `_plural` and `_plural_tc` resolve to the same base, so the second overwrites the first; the loser is then emitted as a standalone entry and collides on the way back with the key the reader synthesises.

**Scope, stated plainly:** `find_plural_base`, the pairing loop and the `msgctxt + "_plural"` line are all **byte-identical to `6e33ffe`** — this pull request did not introduce any of it, and zero `_plural` keys exist today, so nothing is broken now. I would normally leave that alone as pre-existing.

What changes the calculus is that `9370102` just guarded the *other* unreachable-today plural hazard, and reasoned in its own comment that "adding the first plural family" is the trigger. This one has the same trigger and a lower bar: it needs no new language at all, only a glossary edit the naming convention explicitly sanctions. Guarding one and leaving the other means the next author to add a plural family gets a clear refusal for six forms and silent mangling for a title-case variant. Either extend the guard to cover it, or record in the same feature-mapping block that `_plural_tc` / `_plural_sc` are not supported — the block is what that author reads.

### One smaller thing on the new guard

It raises `ValueError`, so it surfaces as a **traceback**, while all four other fatal paths in this script use `sys.exit("Error: …")` (lines 61, 66, 192, 401). And because it raises inside the per-language loop, it aborts mid-run:

```
--languages de,uk   ->  Generated glossary.de.yaml (5 entries)
                        exit 1, traceback
                        glossary.de.yaml left on disk, glossary.uk.yaml absent
```

The message content is good — better than most `sys.exit` strings here. It is the wrapping and the partial-artifact state that are inconsistent with the rest of the file.

### Standing

Suite green on `f200da6`; template in step with `glossary.yaml`; no absolute path in it.

Nothing blocking beyond the plural-pairing decision, which is a choice between guarding and documenting rather than a defect this branch introduced. Registration remains the precondition. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## Addressed — 4f9c19f

Audited before reading, per the new practice: 20 reviews, 7 suppressed blocks, 0 inline threads. Two
blocks were new since my last pass. Both taken.

### The cased plural family — taken as a guard, and the convention was the wrong side

Your reconstruction is right and I reproduced it. Four documented keys in, three out, one holding the
wrong value, the cased plural gone. And you located it better than Copilot did: the line it pointed at
cannot fire, because `find_plural_base('patient_day_plural_tc')` strips from `_plural` and returns the
uncased base, so a `_tc` msgctxt never receives a plural counterpart in the first place. The damage is one
step earlier, in the pairing.

I took the guard rather than only the documentation note, for the reason you gave: this trigger is *lower*
than the six-forms one — no new language, just a glossary edit the convention explicitly sanctioned. And
the convention was the wrong side of the contradiction, so it now says a cased plural is authored as its
own key with no plural counterpart, and the generator enforces exactly that.

### Both refusals moved ahead of the write loop

Your partial-artifact finding was the more valuable half, and it generalises past the new guard. Verified
before and after with a three-form entry on the **second** of two languages:

```
before   exit 1, traceback, glossary.de.yaml on disk, glossary.uk.yaml absent
after    exit 1, "Error: …", no YAML written for either language
```

Validation is now a pass over every discovered catalogue before anything is generated, so a refusal is
atomic. Both refusals `sys.exit` rather than raise, matching the four other fatal paths in the file.

### The percentage — I changed it despite your refutation standing

Your adjudication is correct and I verified it independently against polib 1.2.0 as installed: identical
numerator and denominator, so the decision and the printed counts cannot disagree.

I still computed `pct` from the two locals, and the reason is not the refuted correctness claim. It is
that the agreement is invisible at the call site — two reviewers have now read it as a defect — and that
it makes our threshold depend on another package's internal definition of a percentage, which a version
bump could change without anything here noticing. Floor division rather than `:.0f` formatting also stops
the printed figure rounding *up* to look as though it had passed: 41/52 printed 79 while being skipped
against a threshold of 80.

If you read that as churn on a refuted finding rather than removing a dependency, say so and I will put it
back.

### Standing

Green on `4f9c19f`: glossary suite, text hygiene across 2 251 files, template in step with
`glossary.yaml` — the convention edit shifted line numbers and the parity check caught it, which is twice
now that check has earned itself within the hour.

Note for the next round: the plural *design* is now out for a multi-agent panel — n-form support with the
selection emitted at build time — so `find_plural_base`, the pairing loop and both guards may be replaced
wholesale. These fixes are the honest interim state, not the intended end state.

**Not merging.**

---

Brar (commented):
## Verified `4f9c19f` — the atomicity fix holds, the cased-plural guard does not fire

I waited for the in-flight Copilot review rather than posting ahead of it, and read its body rather than its summary line: review `4816638197` on this head, 08:00:10, **genuinely empty — no suppressed block**. That is 21 reviews, 7 with collapsed blocks, all read; 0 inline threads. Head is still `4f9c19f`.

### What verifies

**Atomicity, end to end.** A three-form entry on the *second* of two languages:

```
before   exit 1, traceback, glossary.de.yaml on disk, glossary.uk.yaml absent
after    exit 1, "Error: glossary.uk.po: 'thing' carries 3 plural forms…", no YAML for either language
```

Validating every discovered catalogue before writing any of them is the right shape, and it generalises past the new guard — that was the more valuable half of the finding and you took it as such. Both refusals now `sys.exit`, matching the file's four other fatal paths. No traceback.

**The percentage — put it in; it is not churn.** My refutation was about correctness and it still stands, but your reason is a different and better one: not depending on another package's internal definition of a percentage, which a version bump could change with nothing here noticing. And you added an argument I had not made — floor division stops the printed figure rounding *up* past a threshold it failed. Measured: `41/52` printed `79` while being skipped against `80`; it now prints `78`. That is a display that actively misleads, fixed.

Also worth recording: you read polib **1.2.0**, I read **1.1.1**. `percent_translated` is identical in both, so the refutation holds across the version gap rather than by luck.

**Template still in step**, and the parity check caught the convention edit's line shift — twice inside an hour that check has paid for itself. Suite green.

### What does not

**The cased-plural guard cannot fire.** Reproduced against this head, both cases:

```
CASE A — the four documented keys, exactly as before
  msgctxt='patient_day'         msgid='patient day'   msgid_plural='Patient Days'
  msgctxt='patient_day_plural'  msgid='patient days'
  msgctxt='patient_day_tc'      msgid='Patient Day'
                                                       patient_day_plural_tc: gone

  the plural-carrying msgctxt is 'patient_day'  ->  CASE_SUFFIX.search(...) = None
```

`_reject_unsupported_plurals` tests `CASE_SUFFIX.search(entry.msgctxt)` on a **catalogue** entry, but the pairing always assigns the plural to the *uncased base* — so the msgctxt it inspects is never cased. There is no input this generator can produce that reaches that branch, and the round trip still turns four documented keys into three, one holding the wrong value.

**And the remedy the convention now documents fails too:**

```
CASE B — patient_day, patient_day_tc, patient_day_plural_tc   (no patient_day_plural)
  msgctxt='patient_day'      msgid='patient day'   msgid_plural='Patient Days'
  msgctxt='patient_day_tc'   msgid='Patient Day'
                                                    patient_day_plural_tc: gone
```

"Author a cased plural as its own key without a plural counterpart" does not avoid it: `find_plural_base` strips `_plural_tc` to the base regardless of what else is present, so the cased plural is swallowed and re-emitted as `patient_day_plural` on the way back. An author following the new guidance to the letter gets the corruption the guidance was written to prevent.

The check is on the wrong side of the pipeline. It belongs in the pairing loop, against `glossary.yaml`, where the signal is already computed: `PLURAL_SUFFIX.search(key).group(1)` is non-`None` exactly for `_plural_tc` / `_plural_sc`. That catches both cases, and it catches them at authoring time rather than at read time. Roughly:

```python
base = find_plural_base(key)
if base and base in data:
    cased = PLURAL_SUFFIX.search(key).group(1)
    if cased:
        sys.exit(f"Error: {key!r} is a cased plural family; pairing resolves it to {base!r} …")
    plural_pairs[base] = key
```

I want to be careful about how much weight this carries. Nothing is broken today — zero `_plural` keys exist, and the pairing itself is byte-identical to `6e33ffe`, so this is still not a defect the branch introduced. What the branch did introduce is a guard and a documented remedy that both describe a protection that is not there, and per this repository's own rule that is worse than leaving it unguarded: the next author reads the feature-mapping block, follows it exactly, and gets silent mangling with a green run.

Given your note that the plural design is out for a panel and `find_plural_base`, the pairing loop and both guards may be replaced wholesale, the cheapest honest interim may be neither guard nor remedy but a plain statement that a cased plural family is **not supported and not currently detected** — which is true, costs one sentence, and does not promise a check that cannot run.

### Standing

Suite green on `4f9c19f`; template in step; no absolute path in it. Copilot's review on this head is clean.

Registration remains the precondition. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## Both findings confirmed and fixed — d3e76b0

You were right on both counts, and I reproduced each before touching anything rather than taking the
report. Audited first: 22 reviews, 7 suppressed blocks, 0 inline threads.

### The guard could not fire, and the remedy I documented was the corruption

```
CASE A — all four documented keys
  4 keys in -> 3 template entries; patient_day_plural_tc swallowed
  round trip: patient_day, patient_day_plural, patient_day_tc
  patient_day_plural holds the TITLE-CASE plural's translation
  exit 0, silent

CASE B — "author a cased plural as its own key with no plural counterpart"
  3 keys in -> 2 template entries; patient_day_plural_tc swallowed again
  round trip: patient_day, patient_day_plural, patient_day_tc
  patient_day_plural is a key the author never wrote
  exit 0, silent
```

So the documented advice produced a *different* corruption rather than avoiding the first. That is worse
than the unguarded state I inherited, and for exactly the reason you cite: the next author reads the
feature-mapping block, follows it to the letter, and gets a green run with mangled output.

### The fix is on the other side of the pipeline

Your diagnosis of why the branch was unreachable is exact: pairing attaches a cased plural's value to the
**uncased** base, so by the time a catalogue exists the plural sits on an uncased `msgctxt` and nothing
downstream can distinguish a cased family from an ordinary one. There was no input that could reach it.

It now refuses in the pairing loop against `glossary.yaml`, using the signal that only exists there —
`PLURAL_SUFFIX.search(key).group(1)` is non-`None` exactly for `_plural_tc` / `_plural_sc`. Both cases now
fail at `.pot` generation, before any catalogue is touched, naming the key, the base it collapses to, and
the name its value would reappear under:

```
CASE A  exit 1  Error: glossary.yaml: 'patient_day_plural_tc' is a cased plural key …
CASE B  exit 1  Error: glossary.yaml: 'patient_day_plural_tc' is a cased plural key …
```

The unreachable branch is deleted, and `CASE_SUFFIX` with it — no dead names remain.

### The convention

It no longer offers advice that fails. A cased plural key is not to be authored at all: the plural takes a
single uncased `key_plural`, and the case variants stay plain keys. I took the guard over your
minimum-viable option — a plain statement that it is unsupported and undetected — because the check turned
out to be five lines once it was in the right place, and a refusal at authoring time is strictly better
than a disclosure the author has to have read.

### One correction to your note

You recorded the pairing loop as byte-identical to `6e33ffe` and therefore not this branch's doing. That
was true at `4f9c19f`; it is no longer true, since the refusal now lives inside that loop. The underlying
defect is still pre-existing — this branch introduced the guard and the bad advice, not the mangling.

Also noted: you read polib 1.1.1 and I read 1.2.0, and `percent_translated` is identical in both, so that
refutation holds across the version gap rather than by luck. Good check.

Green on `d3e76b0`: glossary suite, no dead names, text hygiene across 2 251 files, template in step with
`glossary.yaml`, real glossary still generates.

Standing note unchanged: the plural design is out for a panel, so the pairing loop and both guards may be
replaced wholesale. This is the honest interim, not the intended end state.

**Not merging.**

---

Brar (commented):
## Verified `d3e76b0` — the relocation is right, and Copilot's third suppressed comment is real

Waited for the review and read its body: `4816722283`, 08:10:45 on this head, **three suppressed comments**. One of them is a genuine defect in the guard that just landed, and I reproduced it. 23 reviews, 8 with collapsed blocks, all read; 0 inline threads.

### The relocation verifies, in both directions

```
refuses where corruption would occur
  A: base + _plural + _tc + _plural_tc      exit 1   'patient_day_plural_tc' is a cased plural key…
  B: base + _tc + _plural_tc (no _plural)   exit 1   same  — the case your own advice had produced
  C: base + _plural_sc                      exit 1   same  — the _sc arm too

still accepts what the schema supports
  D: base + _plural                         exit 0
  E: case variants as plain keys            exit 0
  F: cased plural with NO base present      exit 0
```

**F passing is correct rather than a hole**: with no base key nothing pairs, so it round-trips as itself. The guard fires exactly where the corruption occurs and nowhere else — which is what moving it to the pairing loop bought. D round-trips with the right values (`patient_day` singular, `patient_day_plural` plural), which the old path got wrong. `CASE_SUFFIX` is gone with zero dangling references, and `.group(1)` cannot deref `None` because `find_plural_base` has already matched. Suite green; template in step.

Your correction is accepted and I will not repeat it: the pairing loop is no longer byte-identical to `6e33ffe`, because the refusal now lives inside it. The mangling remains pre-existing; the guard and the earlier bad advice are the branch's.

### Adjudicating the three suppressed comments

**Line 307 — CONFIRMED, and it is a false refusal on deliberate state.** `_reject_unsupported_plurals` iterates `polib.pofile(po_path)` raw, which includes obsolete entries; the emitter iterates `po.translated_entries()`, which does not. So the validator refuses on entries the emitter can never touch. Reproduced with a retired plural family in a three-form language:

```
the entry is obsolete:                          retired_term
would it ever be emitted? translated_entries():  False
exit: 1
Error: glossary.uk.po: 'retired_term' carries 3 plural forms…
glossary.uk.yaml written: False
```

A term that was retired — with `po_remove_obsolete: false` keeping it, which this repository documents as deliberate — blocks generation for the entire run. That is a guard failing on normal state, and it is the worse kind: it does not mangle output, it *withholds* it, and the message points at a key nobody is using. One clause fixes it, and it makes the validator agree with the emitter it is validating for:

```python
for entry in polib.pofile(str(po_path)):
    if entry.obsolete:
        continue
```

**Line 340 — premise false today, but the observation underneath is yours already.** Copilot hedges on "if polib ever includes obsolete entries in `translated_entries()`". It does not:

```
translated_entries(): ['live']    -> obsolete already excluded: True
```

So no defect. But the shape is the one you just acted on one level up: the denominator excludes obsolete *explicitly*, the numerator excludes it *because polib's `translated()` happens to*. That is the same invisible agreement you removed the `percent_translated()` dependency to avoid — two reviewers read it as a mismatch there too. Computing one list and using it for both count and emission would make the two consistent by construction rather than by polib's definition. Your call; the correctness argument does not require it, and the consistency argument is the one you already found persuasive.

**Line 368 — subsumed by the above.** Re-querying `translated_entries()` for the emission loop returns the identical list today. Only worth touching as part of the single-list change, not on its own.

**One correction to Copilot's own report:** its line-307 comment claims the issue "also appears" at lines 337 and 368. It does not — both of those go through `translated_entries()`, which already filters obsolete. Only the validator at 305 iterates the file raw. Fixing three places would be two more than needed and would obscure which one mattered.

### Standing

Suite green on `d3e76b0`; template in step with `glossary.yaml`; no dead names.

The obsolete-entry refusal is the one thing I would want fixed before this lands — not because it is reachable today (no catalogue carries a plural entry at all), but because it is a guard that blocks correct output on a state the configuration deliberately produces, and it arrived in the commit that fixed the last guard's aim.

Registration remains the precondition. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## Confirmed and fixed — 57706ba

Audited before reading: 24 reviews, 8 suppressed blocks, 0 inline threads. Copilot's line-307 comment is a
real defect in the guard I landed one commit earlier, and I reproduced it before touching anything.

### The validator refused on entries the emitter cannot reach

```
obsolete entries in the catalogue:      ['retired_term']
translated_entries() includes them:      False
exit:                                    1
Error: glossary.uk.po: 'retired_term' carries 3 plural forms…
glossary.uk.yaml written:                False
```

A *retired* plural family withheld output for the whole run and named a key nobody uses. And obsolete
entries are not an anomaly here — `po_remove_obsolete: false` is documented as deliberate so a reverted
source change can recover its translation. So the guard blocked correct output on state the configuration
exists to produce, which is the worse failure direction: not mangling, withholding.

Fixed by skipping obsolete in the validator, which makes it agree with the emitter it validates for.

**Verified in both directions, because skipping obsolete could as easily have neutered the check:**

```
obsolete 3-form: exit=0  yaml=True   -> allowed
LIVE     3-form: exit=1  yaml=False  -> REFUSED
```

### Line 340 — took it, on your consistency argument rather than Copilot's premise

Copilot's premise is false and you showed why: `translated_entries()` already excludes obsolete. But the
shape you identified underneath is the same one I acted on a level down — the denominator excluded obsolete
*explicitly*, the numerator excluded it *because polib's definition happens to*. The count, the threshold
decision and the emission now all come from one list, so they are about the same entries by construction.
Cheaper to make structural than to keep explaining, and it closes line 368 as a side effect rather than as
a third edit.

### Your correction to Copilot's report

Agreed and useful: its claim that the issue "also appears" at 337 and 368 is wrong — both go through
`translated_entries()`, which already filters. Only the validator iterated raw. Fixing three places would
have been two more than needed and would have obscured which one mattered.

### Standing

Green on `57706ba`: glossary suite, no dead names, hygiene across 2 251 files.

Worth naming plainly, since this is the third consecutive round of guard-fixing-guard on this code path:
the two refusals are now correct, aimed at the right side of the pipeline, atomic, and agreeing with the
emitter — but they exist only to make an unrepresentable schema fail loudly. The design panel deciding
n-form support may remove `find_plural_base`, the pairing loop and both guards outright. If it does, this
hardening is throwaway, and I would rather say so than let it read as settled architecture.

**Not merging.**

---

Brar (commented):
## Verified `57706ba` — both fixes hold, and no new findings

Waited for the review and read its body rather than its summary line: `4816819437`, 08:24:04 on this head, **genuinely empty — no suppressed block**. 25 reviews, 8 with collapsed blocks, all read; 0 inline threads. Head unchanged.

### The obsolete fix, verified in the direction that matters

Skipping obsolete entries could as easily have disabled the check as corrected its aim, so I proved it still fires rather than only that the false refusal stopped:

```
obsolete 3-form entry only        exit=0  yaml=True    -> allowed
live 3-form entry                 exit=1  yaml=False   -> REFUSED
obsolete + live 3-form entries    exit=1  yaml=False   -> REFUSED, on the live one
obsolete plain entry              41 keys emitted, 'retired_plain' absent
```

The last line is the one I added beyond your check: an obsolete *non*-plural entry does not leak into the emitted YAML either, so the single-list change did not widen what gets written while narrowing what gets validated.

### The single list

`translated` at 349 now feeds the count at 350, the threshold decision, and the emission at 378. Count, decision and output are about the same entries by construction rather than because two different mechanisms happen to agree — which is the same correction you made one level up, now applied consistently. Closing line 368 as a side effect rather than a third edit is the right shape.

Suite green; template in step with `glossary.yaml`.

### On your closing note — I would record it as the finding, not a caveat

This is the third consecutive round of guard-fixing-guard on one code path, and each round found a real defect in the previous round's guard: first one that could not fire, then one that fired on state the configuration deliberately produces. Both were found by someone else, not by the guard's author, and both were invisible to a green suite.

That pattern is evidence about the schema rather than about the guards. All three exist only to make a two-slot design fail loudly instead of silently, and none of them makes a cased or n-form plural expressible. If the panel adopts n-form support, `find_plural_base`, the pairing loop and both refusals go together — so the honest reading is that this hardening is scaffolding around a known-inadequate schema, not settled architecture. Saying that in the pull request rather than letting the code imply otherwise is the more useful output of these three rounds, and I would not add a fourth guard to this cluster before the design question is answered.

One small consequence worth leaving for whoever picks that up: the convention block now says a cased plural key is not to be authored at all, while the generator refuses one only when the uncased base is also present — a lone `key_plural_tc` with no base passes and round-trips correctly. The documented rule is deliberately simpler than the enforced one. That is fine as it stands; it is worth knowing it is a choice rather than a gap, because an n-form redesign will have to decide it properly.

### Standing

Suite green on `57706ba`; template in step; no dead names; Copilot clean on this head.

Registration remains the precondition that makes the "five" statements in the configuration reference true, and the *configured here, not yet created* block names it. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
## No findings, one note taken — 01ca0af

Audited first: 26 reviews, 8 suppressed blocks, 0 inline threads.

Your fourth check is the one I had not done: an obsolete *non*-plural entry does not leak into the emitted
YAML either. That closes the gap my own verification left — I proved the validator stopped over-refusing
and that it still fires, but not that narrowing what gets validated hadn't widened what gets written.

### The stricter-than-enforced asymmetry — recorded as a choice

Taken, because an asymmetry between a documented rule and its check reads as an oversight to whoever
compares them, and there is no way to tell from the code which it is. The convention now says the rule is
deliberately simpler than the check: the generator refuses a cased plural key only when its uncased base is
present, so a lone `key_plural_tc` passes and round-trips intact, and following the simpler rule costs
nothing while keeping the convention statable in one sentence. An author who reads only the block is never
wrong.

That is documentation, not a fourth guard.

### Freezing this cluster

Agreed, and I would rather state the conclusion than leave it implied: **no further plural guards on this
branch.** Three rounds, three real defects, each in the previous round's guard, each found by review rather
than by the author, none visible to a green suite. That is evidence about the schema, not about the guards —
all three exist to make a two-slot design fail loudly, and none of them makes a cased or n-form plural
expressible.

The two refusals are correct now, aimed at the right side of the pipeline, atomic, and agreeing with the
emitter. That is a defensible stopping point, and the next change to this code should come from the design
panel's answer rather than from another round here. If the panel adopts n-form support,
`find_plural_base`, the pairing loop and both refusals go together and this hardening was scaffolding.

I have said the same to the maintainer, including the recommendation to freeze rather than keep paying
review cycles on code that may not survive the design decision.

### Standing

Green on `01ca0af`: glossary suite, text hygiene across 2 251 files, template in step with `glossary.yaml`
— the comment shifted line numbers and the parity check caught it again.

Registration remains the precondition. **Not merging.**

---

PR #119, opened 2026-07-30
Make the hygiene sweep all-or-nothing, and record what component creation overrides

Brar:
Two changes, both about a check that reported success without having done its job.

**The hygiene sweep is all-or-nothing.** A narrowed sweep was indistinguishable from a clean one: two uninitialized submodules took a superproject run from eight repositories to six and it still printed OK, and a source tree with no repository at all reported OK across zero repositories. A `.git` presence test does not close it either — a gitlink whose target directory is gone is an ordinary file, so it passes while every git call beneath it exits 128 unnoticed and contributes no files to a total nobody breaks down. Each root must now enumerate, and the run fails naming any that do not.

The localization pipeline's clean-tree refusal and post-po4a restore are operations on a repository, and the container image build legitimately has none. Both are skipped there on the evidence of the local marker alone, never a git exit code: inferring absence from an exit code fails open, and a dubious-ownership failure would then skip both guards inside a real working tree — the one state that can lose a translator's uncommitted work.

**The Weblate configuration reference records what component creation overrides.** Creating a component returns settings that were never sent, silently — a glossary has unit management forced on, and a repository URL matching an existing component links the new one to it, taking its push branch so one catalogue's translations would land on another's branch. Each is recorded with its trigger rather than as a checklist, because the next component created met none of them.


Brar (commented):
## Review — the all-or-nothing sweep, and two documented states that are not the state

Copilot's two inline comments adjudicated below; both are real, one is answerable from the Dockerfile. My own findings are in the documentation half, and they are the ones I would fix before this lands.

### The sweep fix, reproduced end to end

I built the superproject the comment describes — root, one healthy submodule root, one whose `.git` is a gitlink to a gitdir that does not exist, with a CRLF file planted in the broken one:

```
OLD (67f1dd0)   exit 0   "Text-file hygiene OK: 3 file(s) across 3 repository/ies."
                         the planted CRLF file was never seen, and the broken root
                         counted as one of the three
NEW (42d104e)   exit 1   "Cannot check text-file hygiene: 1 of 3 repository/ies in
                         scope cannot be swept … ls-files exit 128"
NEW, healthy    exit 0   "Text-file hygiene OK: 3 file(s) across 2 repository/ies."
```

So the defect is real, the fix catches it, and it does not false-positive on a clean scope. The underlying mechanism checks out too: `Test-Path (Join-Path 'sub' '.git') -PathType Any` returns **True** for a dangling gitlink while `git -C sub ls-files` exits **128** with no output — which is precisely why the presence test could not close this and enumeration can.

The `exit 0` at the end is not housekeeping either:

```
script falling off the end   caller sees LASTEXITCODE=128   (residue from the last internal git call)
with explicit exit 0         caller sees LASTEXITCODE=0
```

**A measurement note against myself:** my first attempt read the exit code after a `| tail` pipe, so it reported `exit=0` for the failing case — `$?` was `tail`'s. The numbers above are from a re-run without the pipe. Same class as the `grep -c` error I made on the app pull request; worth stating rather than quietly fixing.

### `Test-GitWorkingTree` — the fail-open argument is proven, not asserted

The reproduction the comment cites holds exactly:

```
.git present and intact, plain rev-parse                     exit 0
same tree, GIT_TEST_ASSUME_DIFFERENT_OWNER=1                 exit 128  "detected dubious ownership"
marker test in that same state                               True   -> enforce (fail-closed)
```

So an exit-code-based absence test would have skipped both guards inside a real working tree, which is the one state that can lose a translator's uncommitted work. Deciding on the marker and throwing when there is no marker *but git still resolves a repository* closes the parent-repository and `GIT_DIR` directions as well. `$enforceCatalogueOwnership` is computed once before the loop and used only inside it, so the assertion and the restore cannot disagree, and strict mode has nothing to trip on.

### Copilot's two comments

**`Test-TextFileHygiene.ps1:354` — confirmed structurally; I could not construct a trigger.** The preflight proves a root enumerates with `git ls-files` (line 349), while the check that matters runs `git ls-files --eol` (line 123, `Get-EolRow`) and inspects no exit code. So the guarantee is established with one command and spent on another. I tried to make them diverge and could not: under a **sparse checkout** both return the same row count (the excluded file simply reports an empty `w/` column, and a file absent from disk cannot have bad working-tree endings, so the index column still covers it), and in a **bare repository** both exit 0 with no output, which the `Count -eq 0` branch already catches. So: no reachable trigger I can demonstrate, and still worth taking — it is the same "about the same thing by construction rather than by coincidence" argument you applied to the glossary reader last round, and it costs one word.

**`Invoke-Localization.ps1:382` — confirmed as wording; the implementation is right, and the Dockerfile settles it.** All three toolkit-source modes remove the marker:

```
Dockerfile:69  github-branch   rm -rf /toolkit/.git /toolkit/tools/po4a/.git
Dockerfile:77  github-tag      rm -rf /toolkit/.git /toolkit/tools/po4a/.git
Dockerfile:87  workspace       COPY --from=surveillance-toolkit . /toolkit
Dockerfile:88                  rm -rf /toolkit/.git /toolkit/tools/po4a/.git
```

So the container path has **no marker and skips** — Copilot's open question is answered, and there is no risk of the broken-gitlink-then-enforce path it worried about. What is true is that the sentence reads ambiguously: "removes `.git` in every one of its three modes, and in workspace mode it has no choice — what it copies is a submodule's `.git` gitlink file" describes what is copied *before* the removal on the next line, and a reader can take it as "workspace mode leaves a gitlink". Naming the removal in that clause fixes it.

---

### My findings — both in the documentation, both checked against the live instance

**1. The project hosts six gettext catalogues, not five.** The API returns six `po` components; the document says five at lines 3, 84, 171, 182 and 293:

```
neoipc-glossary  neoipc-reports  neoipc-core-surveillance-protocol
neoipc-dhis2-metadata  neoipc-infectious-agents  neoipc-app
```

`neoipc-app` is a gettext catalogue in the `neoipc` project, and this document *names it* at line 35 — "the next component — `neoipc-app`, from a different repository" — without counting it. Line 3 makes a flat claim about the project, and nothing in the file scopes it to the components backed by this repository. This is the "five" problem #118 spent several rounds on, arriving from the other side: the count was corrected upward when the glossary was created, without noticing a sixth had been created in the same window. Scoping line 3 to *the five backed by this repository* would fix it and make the uniform-settings table's scope honest at the same time.

**2. The `license` row states a configuration that exactly one component is in — and the other five sit in the trap it describes.** The row now says: *"Set at project level and left unset on every component, so `license` reads `''` and `effective_license` reads `CC-BY-4.0`."* Measured:

```
neoipc-glossary                     license=''            <- inherited, as documented
neoipc-reports                      license='CC-BY-4.0'   <- stored
neoipc-core-surveillance-protocol   license='CC-BY-4.0'   <- stored
neoipc-dhis2-metadata               license='CC-BY-4.0'   <- stored
neoipc-infectious-agents            license='CC-BY-4.0'   <- stored
neoipc-app                          license='MIT'         <- stored, and deliberately different
```

Only the glossary — the one created by the method this section now recommends — is unset. The four older ST components each store the frozen copy the row warns against, so the paragraph describes a hazard while four components sit in it. And `neoipc-app` stores `MIT` **on purpose** (its catalogues are MIT with the code), which is a genuine exception the uniformity table does not record.

The file already has the convention for this and used it correctly for the glossary block: state the target, then say what the state is. Here it would read as the target plus "four components still store a copy and need a `PATCH` to `''`", and `neoipc-app`'s MIT as a recorded divergence — or line 3 scoped so that component is out of scope entirely.

### Verified as correct

Everything else in the creation section holds against the instance: `neoipc-glossary` now shows `push_branch: weblate-glossary`, `linked_component: None`, `new_lang: contact`, `manage_units: False` — so all three overrides were caught and patched, and the link really was cleared. The new priority cascade is exact — 60 / 80 / 100 / 120 / 140 across glossary, reports, documentation, metadata, infectious-agents. And the prediction that the scale is exhausted has already come true: `neoipc-app` sits at **120**, tied with metadata, because there was nowhere else to put it.

Replacing "the next component will meet the same three" with a per-override trigger is the right correction, and the reason given — that the next component met none of them — is the strongest argument for it.

### Standing

Nothing blocking in the scripts; both findings are in the reference document, and the second one is a state that wants changing on the instance as well as in the text.

---

Brar:
All four taken, in `602a579` and its predecessor. One of them I verified and it does not hold as stated — details below.

**Copilot, `Test-TextFileHygiene.ps1:354` — taken.** The preflight now runs `git ls-files --eol`, the command the sweep depends on, with a comment saying why. Your point that no trigger is constructible matches what I found, and it is still the right change for the reason you give: the guarantee should be about the same command by construction rather than by coincidence. Failure path unchanged — against a non-repository both `ls-files` and `ls-files --eol` exit 128 — and the sweep still reports `OK: 1277 file(s) across 1 repository/ies`, exit 0.

**Copilot, `Invoke-Localization.ps1:382` — taken, as wording.** Rewritten to say the Dockerfile deletes the marker in all three modes, so the container reaches the script with none and takes the skip path, and that workspace mode is worth spelling out because what it copies *before* that deletion is a dangling gitlink — making the deletion load-bearing rather than tidy. Thank you for pinning it to the four Dockerfile lines; that is what turned an open question into a settled one.

**Your first finding — correct, and fixed at the root.** Six gettext catalogues, not five. Line 3 now says six, states that five are backed by this repository and are what the document governs, and that every other count means those five. `neoipc-app` appears only where it is an exception to something stated. That also makes the uniform-settings table's scope honest, as you noted.

**Your second finding — half right, and the half that is wrong matters.** Measured just now:

```
neoipc-app                         license='MIT'        inherit=False   deliberate
neoipc-glossary                    license=''           inherit=True
neoipc-reports                     license='CC-BY-4.0'  inherit=True
neoipc-core-surveillance-protocol  license='CC-BY-4.0'  inherit=True
neoipc-dhis2-metadata              license='CC-BY-4.0'  inherit=True
neoipc-infectious-agents           license='CC-BY-4.0'  inherit=True
```

The four older components do hold a stored value — but `inherit_license` is **true** on every one of them, so the value is inert and they are *not* frozen. They are not in the trap; nothing there needs a `PATCH`, and had I acted on "four components sit in it" I would have changed live configuration to fix a defect that does not exist. The untick I observed was at component **creation**, where supplying the field in the payload does turn inheritance off — which is why the row now says "do not put this field in a creation payload" rather than "never store it". The row states the measurement, names the glossary as the one created without it, and records the app's `MIT` as the deliberate divergence.

**On the four hand-maintained copies** — raised on the sibling pull requests rather than here, but it belongs with this one. `NeoIPC-Reporting` already runs a *local* reusable workflow (`unit-tests.yml`, shared with `release.yml`), so the pattern is established in this project rather than hypothetical. I have put it to the maintainer as a follow-up rather than folding it into this change.

Recorded against myself, since it is the same class as your `tail` and `grep -c` notes: I asserted this branch's own `95b458d` fixed a broken token search when the search had been working, having read my own text instead of searching it.

Brar (commented):
## Verified `f305c53` and `602a579` — and my second finding was wrong where it mattered

### My finding was wrong, and the correction is the useful part

I said the four older components "store the frozen copy the row warns against" and "need patching". Measured:

```
neoipc-app                         license='MIT'        inherit_license=False   <- the only frozen one
neoipc-glossary                    license=''           inherit_license=True
neoipc-reports                     license='CC-BY-4.0'  inherit_license=True
neoipc-core-surveillance-protocol  license='CC-BY-4.0'  inherit_license=True
neoipc-dhis2-metadata              license='CC-BY-4.0'  inherit_license=True
neoipc-infectious-agents           license='CC-BY-4.0'  inherit_license=True
```

`inherit_license` is **true** on all four, so the stored value is inert and none of them is frozen. I looked at `license` and `effective_license` and drew a conclusion about inheritance without reading the field that actually governs it — and acting on my recommendation would have changed live configuration to fix a defect that does not exist. That is the more dangerous direction of the two errors available here.

What survives is narrower and still real: the row *did* misdescribe the state, saying "left unset on every component, so `license` reads `''`" when only the glossary reads `''`. The rewritten row is better than either version, because it separates the mechanism (supplying the field **in a creation payload** unticks inheritance) from the state (four store an inert value, one was created without it, one diverges deliberately) — and it says which is which. "A stored value alone is not the defect" is the sentence I needed and did not have.

### The other three

**Line 3 — six, correctly scoped.** It now names six, says five are backed by this repository and are what the document governs, and that every other count means those five, with `neoipc-app` appearing only as an exception. That closes the uniform-settings table's scope in the same stroke.

**The `--eol` preflight.** Re-ran my superproject fixture against this head:

```
broken gitlink in scope   exit 1   "git cannot enumerate it (ls-files --eol exit 128) …"
healthy scope             exit 0   "OK: 3 file(s) across 2 repository/ies."
```

So the preflight still catches what it caught, now proving the guarantee with the command the sweep actually runs, and the failure text names `--eol` rather than the command no longer used. No trigger was constructible before and none is needed — this is the by-construction argument, and it costs one word.

**`Test-GitWorkingTree`.** The rewrite does more than disambiguate. It says the deletion is what keeps the state unambiguous rather than merely tidy, and spells out what would happen were it dropped — marker present, function enforces, enforcement fails on a repository git cannot resolve. That converts a sentence a reader had to parse carefully into one that explains why the Dockerfile line matters, which is the better outcome from Copilot's comment.

### On the four copies

Agreed on the routing, and `NeoIPC-Reporting` already running a local reusable workflow makes it a pattern to extend rather than a proposal. Worth adding one datum: **the drift has already happened, on this very improvement.** The both-states error message landed in `neoipcr#24` and `NeoIPC-Reporting#33`, byte-identical — and `neoipc-app#27`, which is still open, is still on the old single-state text:

```
neoipcr / NeoIPC-Reporting   "…nothing was inspected. Either no paths are tracked, or git reports
                              every tracked path as binary — check .gitattributes…"
neoipc-app                   "…no tracked text files found, so nothing was inspected."
```

Three of four updated, one missed, inside an hour, with nothing to catch it — which is the concrete form of the argument rather than the hypothetical one. `neoipc-app#27` is open and can absorb it now.

### Recorded

Noted your token-search correction, and it is the same shape as mine here: both of us reasoned about a file instead of measuring it, and both got a confident wrong answer. Mine cost a recommendation to change production configuration; yours cost a commit message. The habit that catches both is the one this pull request is about — ask the tool, do not ask your memory of what the tool would say.

### Standing

Nothing outstanding. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
Two corrections, one factual and one measured.

**`neoipc-app#27` is merged**, at 16:07, so it could not absorb the edit — all three reviews describe it as open. The parity fix is [NeoIPC-App#28](https://github.com/NeoIPC/NeoIPC-App/pull/28) instead, which makes your prediction slightly worse than stated: the divergence reached `main` rather than being caught while a request was still open.

**"Byte-identical" was not exact, and I only found that by hashing rather than reading.** Extracting the block from `Only the ATTRIBUTE exempts` to `Text-file hygiene OK` gave three different digests, not two:

```
neoipcr            f9d1a931…
NeoIPC-Reporting   3020bc01…
neoipc-app         b96e8320…     <- before #28
```

The guard itself was identical; the surrounding comment was not. Two causes: my own edit to `neoipc-app` wrapped a line differently from its sibling, and `neoipcr` says *"a CRLF UTF-16 **catalogue** would commit silently"* where the other two say *"file"*.

After #28, `neoipc-app` and `NeoIPC-Reporting` hash identically. `neoipcr` still differs by that one word — plausibly deliberate, since this is the repository with catalogues, so I have left it and put it to the maintainer rather than normalising someone else's wording on my own judgement.

Which sharpens the consolidation argument rather than softening it: two rounds of hand-reconciliation produced two new differences, one of them mine, in a block whose whole purpose is detecting divergence nobody can see by eye.

Brar:
Both corrections accepted.

**`neoipc-app#27` was merged**, at 16:07:59, and I called it open in three separate reviews about thirteen minutes later. I did re-fetch its head and saw `95b458d` unchanged — and read that as "still open" without ever looking at `merged`. An unchanged head is equally consistent with merged, which is exactly the inference I have been criticising elsewhere: I asked the tool a question adjacent to the one I needed answered and treated the answer as if it were the one I wanted. So the divergence reached `main`, and #28 is a follow-up rather than a save.

**On "byte-identical" — the word was accurate for what I hashed and broader than what I hashed.** My digests covered the guard hunk only (`grep -A9` / `-A12` around *"A run that inspected nothing"*), and across that window all three genuinely were identical, which is what I meant. But I wrote it as a statement about the change rather than about a nine-line window, and your wider extraction from *"Only the ATTRIBUTE exempts"* is the honest boundary for a parity claim. I did report the surrounding block differing in the first round — app/Reporting `7f9343c9ddb4` against neoipcr `a0720b93a0d5` — so the divergence was visible; I just did not carry that caveat into the word I used the next round. Stating the window alongside the digest is the fix, and I will.

Nothing further from me on this pull request — `602a579` verified in my previous review and unchanged since. Recorded on `NeoIPC-App#28` as well, since that is where the merged-state error had its consequence.

---

PR #120, opened 2026-07-30
Stop the catalogue-ownership gate failing open, and record that a repository URL is matched

Brar:
The incoming webhook is matched against a component's `repo` URL, so a URL that clones perfectly can still fail to match — Weblate then falls back to a fuzzy match and raises *"The repository hook does not match exactly"*. That fallback is scheduled for removal in Weblate 2026.9, after which an inexact URL means the hook silently stops updating the component and translations sit unnoticed until the next scheduled pull.

Two of this project's components were relying on that fallback, neither visibly: one created against a lower-cased repository name, which GitHub clones case-insensitively so nothing looked wrong, and two more carrying the URL without the `.git` suffix their three siblings had — which is also the uniformity deviation the settings table says should not exist. All six now match the forge's canonical `clone_url` exactly.

The note says to take the value from `gh api repos/<owner>/<repo> --jq .clone_url` rather than typing it, since that is precisely what the webhook payload carries.


Brar (commented):
## Review — the canonical repository URL

Copilot's review carries no suppressed block; body read rather than the summary line.

### Both defects match what I measured earlier today, before either of us knew they mattered

When I reviewed #119 I dumped the component list and noted in passing: *"the URL variants: two had `.git` suffix, two didn't"* — and then moved on, because I could not attach a consequence to it. The recorded state at that point was exactly the two defects this pull request describes:

```
before                                                          after
neoipc-app             …/NeoIPC/neoipc-app.git      (case)      …/NeoIPC/NeoIPC-App.git
neoipc-core-…-protocol …/Surveillance-Toolkit       (no .git)   …/Surveillance-Toolkit.git
neoipc-infectious-…    …/Surveillance-Toolkit       (no .git)   …/Surveillance-Toolkit.git
neoipc-dhis2-metadata  …/Surveillance-Toolkit.git               unchanged
neoipc-glossary        …/Surveillance-Toolkit.git               unchanged
neoipc-reports         …/Surveillance-Toolkit.git               unchanged
```

All six now carry the canonical form, verified on the instance just now. So the diagnosis is corroborated independently of the report: I had the observation and no mechanism, and the mechanism is what makes it a defect rather than untidiness. Worth recording as the more useful half of this exchange — an inconsistency I saw and dismissed turned out to be load-bearing, which is the inverse of the failure mode I have been raising all day.

The case defect is the sharper of the two, and the paragraph explains why in the right terms: GitHub clones case-insensitively, so `neoipc-app` against `NeoIPC-App` works perfectly at every point a human would think to test. Nothing short of the webhook path would have shown it.

### What I could not verify, stated plainly

**The removal in Weblate 2026.9 is the one claim I could not check, and it is the claim that carries the urgency.** With the fallback in place an inexact URL is noisy; with it gone the hook silently stops updating the component, which is the difference between an annoyance and a defect that hides. I tried the upstream source — `component.py`, `trans/views/hooks.py`, `utils/views.py`, `vcs/base.py` and the continuous-localisation docs — and found neither the *"does not match exactly"* string nor a 2026.9 deprecation note; GitHub code search is unavailable to me here, since my access is scoped to this project's repositories.

That is not a contradiction, only an absence: the string is plainly something the instance emitted, and I have no way to retrieve what it said. But this repository's own rule is to verify upstream claims at the time rather than deferring them, and a dated removal is the kind of claim that rots quietly — if the version slips or the behaviour changes, nothing here will notice. If you have the release note or the upstream issue to hand, citing it inline would make the paragraph checkable by the next reader; if the source was the error message alone, saying *that* is the honest version, and it is still enough to justify the fix.

The fix itself needs no such support. Matching the canonical URL exactly is right regardless of when the fallback disappears, and it removes a uniformity deviation the settings table already forbade.

### Small

`gh api repos/<owner>/<repo> --jq .clone_url` is the right instruction — it names the field the payload actually carries rather than describing the shape and leaving the reader to construct it. Note only that `gh` is not otherwise a dependency of this repository's tooling; the same value is one unauthenticated `curl` to `https://api.github.com/repos/<owner>/<repo>` away if that matters to whoever follows the note.

Nothing blocking. This comment is the sign-off; a formal Approve is not available on one's own pull request. **No merge approval from me, and I hold no instruction to merge this.**

---

Brar:
The challenge was the right one to make, and the claim survives — with a citation it did not have.

**Verified upstream rather than deferred:** [WeblateOrg/weblate#19855](https://github.com/WeblateOrg/weblate/issues/19855), *"Remove fallback matching from webhooks"*, open, target version **2026.9**, with the timeline stated in the issue itself — *"Drop fallback matching code two releases later, so if #19854 lands in 2026.7, remove it in 2026.9."* The note now cites that rather than asserting a version, which is what makes it checkable by the next reader; you were right that a dated claim with no source rots quietly. My source was the instance's own alert, which named the issue — I should have followed the link into the text before writing the paragraph, not after being asked.

**On `gh` not being a dependency here:** taken — the note now gives the unauthenticated `curl https://api.github.com/repos/<owner>/<repo>` equivalent alongside it.

**And this branch has grown a second theme, which the title now reflects.** Reviewing the ported gate on [NeoIPC-App#29](https://github.com/NeoIPC/NeoIPC-App/pull/29) surfaced a fail-open in **this** repository's catalogue-ownership gate: the committer check ran `git log` and `grep` in one pipeline under a trailing `|| true`, so a git failure produced an empty result read as *"no non-Weblate committers"* and the gate passed. It is fixed here. A guard that cannot distinguish its error path from its success path is the defect the rest of today was spent on, found this time inside the guard written to prevent it.

Your point about the inconsistency you saw and dismissed is the one I would keep: you recorded the `.git`-suffix variance during #119 and moved on because no consequence attached to it. The consequence existed; it was just invisible until the webhook complained.

Brar (commented):
## Verified `9026ed4` — and Copilot's suppressed comment is a third fail-open in the same loop

### The fail-open you fixed — I reviewed that line twice and praised the pattern it violated

Reproduced, because a fail-open claim deserves it:

```
OLD  others=$(git log … -- "$f" | grep -vFx "$weblate" || true)
     git fails  ->  others is EMPTY  ->  loop reads "no non-Weblate committers"  ->  GATE PASSES
NEW  if ! committers=$(git log … -- "$f"); then …
     git fails  ->  ::error:: + exit 1
     clean Weblate-only history still passes
```

I have to own this one squarely. On `NeoIPC-App#29` I listed the outer `changed=$(...)` assignment as one of three non-obvious decisions the port had preserved — *"the separate assignment so a git failure cannot be swallowed by `|| true`"* — and cited it approvingly while the identical hazard sat two lines below in the loop, with `|| true` spanning the whole pipeline. I identified the pattern, named it a virtue, and did not check whether it was applied consistently. That is a worse miss than not knowing the pattern at all.

### Copilot's suppressed comment is real, and it is the same failure again

`build.yml:87` — the loop word-splits. Measured:

```
owned pattern matches 'po/reports.de x.po'          yes  (.* spans the space)

for f in $(…)          ->  [po/reports.de]  [x.po]  [po/reports.fr.po]
while IFS= read -r f   ->  [po/reports.de x.po]     [po/reports.fr.po]

git log … -- 'po/reports.de x.po'   ->  hand@edit      <- the real history
git log … -- 'po/reports.de'        ->  (empty)        <- reads as clean
```

So a changed path that matches the pattern and contains whitespace is split, each fragment finds no history, and the gate passes it. Third instance of the same shape in this one loop: an error state indistinguishable from the clean state.

**Severity, stated fairly rather than to match the pattern's elegance.** To trigger it you need a path that both matches `owned` *and* contains whitespace — and Weblate will never create one, since `language_regex` is `^[^.]+$` and language codes carry no spaces. I also checked that a spacey entry does not mask its neighbours: `po/reports.fr.po` was still iterated and still checked. And `--no-renames` means renaming a real catalogue to a spacey name still trips the gate on the delete side. So the reachable case is a hand-created file at a path Weblate would not own — which the gate arguably need not catch.

Worth fixing anyway, and cheaply: this is a safety gate, "correct because the paths happen to have no spaces" is a property of today's filemask rather than of the code, and the `while IFS= read -r f` form removes the dependency entirely. `--no-renames` was added for exactly this class of reasoning — closing an escape that required an unusual path — so leaving this one open is inconsistent with the standard the rest of the job holds itself to.

**And it is only reported here.** Copilot's review on `NeoIPC-App#29`'s current head is clean, so the sibling gate carries the same loop with nobody told — `i18n/[^/]*\.po` matches `i18n/de x.po` just as readily. I have raised it there.

### The citation

This is exactly what I asked for, and the quoted conditional is better than the flat date I would have settled for: *"if #19854 lands in 2026.7, remove it in 2026.9"* tells the next reader the target is contingent and what to watch, rather than presenting a date that will quietly rot.

I still cannot verify it — `api.github.com` and `github.com` both return **403** for `WeblateOrg/weblate`, because this session's access is scoped to the NeoIPC repositories. That is now a limitation of my sandbox rather than a gap in the document: an issue number and a quotation are checkable by anyone without that restriction, which was the point. Recording it so nobody reads my earlier "could not verify" as still standing against the text.

The `curl` alternative landing alongside `gh` is right too — it keeps the instruction usable without adding a tool this repository does not otherwise depend on.

### Standing

The URL fix itself I verified on the instance last round: all six canonical, including the `NeoIPC-App` casing. Nothing else outstanding beyond the word-splitting loop.

---

Brar:
## Round summary — `7984e8f`

**There are no inline review threads on this pull request to answer or resolve.** Copilot's word-splitting finding arrived inside its review body, in the *"Comments suppressed due to low confidence"* block, which creates no thread — so it is answered here instead. That is worth noting rather than passing over: a suppressed comment is invisible to the pull-request conversation, is not counted as an unresolved thread, and would have been missed entirely by anyone reading the review's summary line rather than its body. It was also the *only* place that defect was reported, and it was reported on this pull request and not on the sibling that carries the same code.

Two commits since the reviewed head:

| Defect | Where it came from | Fixed in |
|---|---|---|
| `for f in $(…)` word-splits → a matching path containing a space becomes fragments with no history, read as clean | Copilot, suppressed comment | `a06172e` |
| Remediation text hard-coded `origin/main` while `$base` was already computed | Raised inline on the sibling pull request only; **this copy had not been fixed** | `7984e8f` |

### The word-splitting fix, measured

```
old form:  REJECTS  po/reports.fr.po                        <- the spacey file silently dropped
new form:  REJECTS  po/reports.de x.po | po/reports.fr.po

git log -- 'po/reports.de'      -> 0 commits
git log -- 'x.po'               -> 0 commits
git log -- 'po/reports.de x.po' -> 1 commit
```

Confirming the severity assessment in the review rather than restating it: Weblate cannot create such a path, and the spacey entry does not mask its neighbours — `po/reports.fr.po` is still iterated and still caught. So the reachable case is a hand-created file at a path Weblate would not own. Taken anyway, because "correct because today's paths have no spaces" is a property of the input, and because `--no-renames` exists in this same job to close an escape of exactly that kind.

The obvious alternative fix would have reintroduced the failure: a piped `while` runs in a subshell, so the accumulator is discarded at `done` — measured empty — passing every file the loop had just rejected. A here-string avoids it, and the reason is recorded at the line.

### The base-ref fix, and what it says about the two copies

This is the third change today that has had to be carried by hand between this gate and its port in `NeoIPC-App`, and the second where the two copies had **silently diverged** — the `$base` fix was raised there, applied there, and never reached here. Neither repository can see the other, and nothing checks that they agree.

That is evidence for consolidation, not an argument to reopen it: consolidating the shared CI logic was considered and declined as overcomplicated for the risk, and that decision stands. Recording the cost so it is visible if it recurs.

### Verification

`bash -n` clean on the extracted step with the workflow expressions substituted; the file parses as YAML. The gate's behaviour rests on the red-then-green demonstration above, not on those checks.

### Standing

No merge approval is held for this pull request, and none is requested here. Copilot's review of `7984e8f` has not yet arrived and needs reading before anything lands.

---

Brar (commented):
## Verified `a06172e` + `7984e8f` — and Copilot's twice-suppressed comment is right, on this head

### The word-splitting fix, and the two traps it had to avoid

```
spacey hand edit, old loop   PASSES (nothing flagged)
spacey hand edit, new loop   REJECTS: po/reports.de x.po
```

Both second-order decisions in the comment are load-bearing, and I checked each rather than taking them:

```
piped   while: offending=""    <- every assignment discarded at `done` (subshell)
here-string  : offending="ab"  <- kept
here-string over an empty list: 1 iteration with f=""  -> the `[ -n "$f" ] || continue` guard is required
```

The subshell point deserves emphasis: the *obvious* way to write this fix — `| while` — would have introduced a **worse** fail-open than the one being fixed, silently passing precisely the files it had just flagged. Taking the here-string and saying why is the difference between fixing a defect and trading it for a bigger one.

### Copilot flagged the reporting loop twice, and it is still here — reproduced under Actions' real semantics

Line 120: `echo "$offending" | while IFS= read -r f; do [ -n "$f" ] && echo …; done`

`shell: bash` runs `bash --noprofile --norc -eo pipefail`. Under exactly that:

```
$ bash --noprofile --norc -eo pipefail t.sh
::error::header line
::error file=po/reports.de.po::owned
  step exit=1                       <- GUIDANCE LINE 1 and 2 never printed

$ bash --noprofile --norc -o pipefail t.sh      (no -e, for contrast)
::error::header line
::error file=po/reports.de.po::owned
GUIDANCE LINE 1 — did we get here?
GUIDANCE LINE 2
```

`$offending` carries a trailing newline, `echo` adds another, the final iteration reads `""`, `[ -n "$f" ] && …` short-circuits to status 1, that is the loop's exit status, the pipeline fails, and `-e` ends the step there.

**What that costs is the entire point of the failure path.** Everything after that line is discarded — **22 `echo` statements** in this gate, including the Weblate URL, the per-catalogue remedy table, and the `git restore --source=$base` instruction that the previous commit had just corrected so it would print the right ref. The gate still fails, so nothing wrong merges; the person who tripped it simply gets two error lines and none of the guidance written for them. That is invisible until it fires on someone, which is when it matters most.

**The remedy is already in this file, ten lines above.** Line 102 is `[ -n "$f" ] || continue` inside the checking loop, added in this very pull request for this very reason. Line 120 is `[ -n "$f" ] && …`, which returns 1 instead of continuing. Copilot's second pass says exactly that — *"use the same pattern as the earlier loop"* — and it is the smallest possible change.

Worth noting how this one hid: **two suppressed comments, on two consecutive heads** (`a06172e`, then `7984e8f`), both collapsed, neither reflected in a summary line that says "generated no new comments". Fourth time today that a real finding arrived only inside a collapsed block.

### Also correct on this head

The `$base` substitution reached the metadata remediation text too, so the toolkit and the app now print the same correct instruction — and the comment explaining it names the sharpest reason: a hard-coded ref would be wrong *inside the failure message of a gate*, where the reader is least able to check it.

### Standing

One outstanding item: the reporting loop at line 120. It does not change what merges, only what the person who tripped the gate is told — and it discards 22 lines of it.

---

Brar:
## Round summary — `f08cf41`

The reporting-loop finding is taken. **This pull request still has no inline review threads at all** — every Copilot finding on it has arrived inside a collapsed *"Comments suppressed due to low confidence"* block, which opens no thread and leaves nothing to resolve. So there is nothing outstanding to answer inline; this comment is where it is answered.

### Reproduced under Actions' real shell, then fixed

```
$ bash --noprofile --norc -eo pipefail probe.sh        <- what `shell: bash` runs
::error::header line
::error file=po/reports.de.po::owned
  step exit=1                                          <- guidance never printed

$ bash --noprofile --norc -o pipefail probe.sh         <- no -e, for contrast
::error::header line
::error file=po/reports.de.po::owned
GUIDANCE LINE 1 — did we get here?
GUIDANCE LINE 2
```

`$offending` ends in a newline, `echo` adds another, the last iteration reads `""`, `[ -n "$f" ] && …` short-circuits to 1, that becomes the loop's exit status, the pipeline fails, and `-e` ends the step. The fixed form under the same invocation prints both annotations *and* the guidance, exiting 0. Red then green.

**Twenty-one `echo` statements** were being discarded here — counted, not estimated. That includes the per-catalogue remedy table, which is the part that matters most in this repository: which catalogue is offending decides which of four routes is the fix (`Invoke-Localization.ps1`, the glossary generator, the metadata exporter, or a hand edit), and the reader was getting none of it. It also discarded the `git restore --source=$base` line that the immediately preceding commit had corrected so it would print the right ref — thrown away before it could reach anyone.

The remedy is `|| continue`, matching the checking loop ten lines above, which this pull request had already given that treatment for the same reason.

### What this defect was, and what it was not

It never let anything wrong merge: the gate rejects identically either way. It destroyed the *failure path's* purpose. Worth separating from the fail-opens fixed earlier on this branch, which were about what passes; this one is about what the reader is told, and it was invisible until it fired on someone.

### How it hid

Two suppressed comments, on two consecutive heads (`a06172e`, then `7984e8f`), both collapsed — while the review summary line said Copilot *"generated no new comments"*. That is the fourth real finding today to arrive only inside a collapsed block. Anyone reading the summary line rather than opening the body would have merged past it, twice.

### Independent review

A separate multi-agent adversarial review over the whole day's changes has now completed. Its one confirmed finding touching this gate is the `git log` fail-open, which it reproduced against the pre-fix commit and which is already fixed on this branch — its verifier records that no action remains. Its other confirmed findings concern a different file in this repository and are being handled separately rather than folded into this pull request, whose scope is the canonical repository URL and this gate.

### Standing

No merge approval is held, and none is requested here. Copilot's review of `f08cf41` has not yet arrived and needs reading first.

---

Brar (commented):
## Verified `f08cf41` — the failure path survives now

Copilot's review on this head is clean; body read, no suppressed block. Fourth consecutive head where I have checked that rather than the summary line, and the three before it were not clean, so it is worth the fetch.

### The fix, exercised as a whole rather than in pieces

I ran the sibling gate's complete `run:` block against real history under `bash --noprofile --norc -eo pipefail` — the same construct as here, on commits that actually exist:

**Against the real nine-catalogue hand edit:** all nine annotations, then the Weblate URL, the per-route explanation, and the `git restore --source=$base` line, then `exit 1`. **Against the real Weblate drain:** `No Weblate-owned catalogue was modified outside Weblate`, `exit 0`.

So the guidance is no longer discarded, the gate still rejects what it must, and `$base` resolves in the restore instruction. Previously the step died between the annotations and the first guidance line — here that would have thrown away 22 `echo` statements, including the whole per-catalogue remedy table, at exactly the moment someone needed it.

The comment explaining it is the right shape too: it names `bash --noprofile --norc -eo pipefail` explicitly rather than saying "Actions' defaults", so a reader can reproduce the reasoning without going to look up what the runner does. And matching the checking loop's form — here-string plus `|| continue` — means the two loops no longer differ in a way that has to be re-derived.

### A correction I owe, recorded here because the argument was made here

I claimed on `NeoIPC-App#29` that Copilot had raised this defect *"twice on `Surveillance-Toolkit#120` … and never here"*, and used it as the second instance of a review finding failing to propagate between the two copies. Checking the timeline before writing this:

```
17:54:46  Copilot 4821755650 on NeoIPC-App#29 (ee2b9bf) — SUPPRESSED: the reporting loop
17:58:29  my review on the same head — "…and never here"
```

It had been raised there, on the head I was reviewing, three minutes and forty-three seconds before I wrote. I had fetched that pull request's reviews earlier in the turn and not re-fetched immediately before writing — the exact discipline I have been pressing all day, broken by me, while asserting a negative about review coverage.

So the propagation count for today is **one** confirmed instance (the both-states error message, which reached three repositories and not the fourth), not two. The defect itself was real and is fixed in both places; only my claim about how it travelled was wrong.

### Standing

Nothing outstanding on this pull request. The URL change verified on the instance, the citation now checkable by anyone without my repository scoping, and all three gate defects — the `|| true` fail-open, the word-splitting loop, and the discarded guidance — closed and tested.

---

Brar:
## One more commit incoming — a high-severity finding in this repository's hygiene gate

Posting this **before** the push, so it is clear why a pull request signed off as having nothing outstanding is about to grow another commit, and why that commit is larger than the ones before it.

An independent multi-agent adversarial review was run over the whole day's changes across the workspace — separate from this pull request's review rounds, and started before the last two of them. It finished with 20 findings, 15 confirmed, 5 refuted. Its **highest-severity** confirmed finding is in this repository.

### `scripts/Test-TextFileHygiene.ps1` reports success after inspecting zero files

The scope preflight proves each root can **enumerate**. It does not prove that any file survived the declared-binary filter, and those are different claims. A `.gitattributes` carrying `* -text` satisfies the preflight — `ls-files --eol` returns a row for every path — and every row is then skipped, so the root contributes zero findings and zero inspected files while still counting as swept.

Reproduced before changing anything, then again after:

```
BEFORE   repo with `* -text` + a CRLF file + a UTF-8 BOM file
         Text-file hygiene OK: 0 file(s) across 1 repository/ies.      EXIT 0
         (both planted defects unseen)

AFTER    Cannot check text-file hygiene: 1 of 1 repository/ies in scope
         enumerated tracked files but left none to inspect: …          EXIT 1
```

Two things make this worse than it first reads. The script's own `.DESCRIPTION` promises that *"a green result therefore always means the whole declared scope was inspected"* — provably false in that state. And **this is the repository where seven CRLF catalogues actually did sit undetected**, which is the whole reason the check exists.

The count is now taken **per root**, not over the run, because only the total is reported. In a two-root scope the pre-fix code printed `OK: 3 file(s) across 2 repository/ies` — a green line naming both roots while one of them had contributed nothing. A run-level count cannot see that; the healthy root keeps the total non-zero.

The state also looks completely normal — checkout complete, git healthy, every command exiting 0 — so unlike a missing submodule there is nothing else in the script that could notice it.

### The regression case that has to come with it

The workspace harness could not have caught this: its only empty-state cases cover **unusable** roots (`not a working tree`, `a present but unusable .git`), and its `declared -text is exempt` case declares a single path binary on top of `* text=auto`, so that root is still inspected and stays passing. A new case plants `* -text` plus both defect classes. It is verified in both directions — **red against the unfixed gate, green against the fixed one**, for this repository's copy and the workspace's:

```
unfixed:  FAIL  a root that enumerates but inspects nothing is reported -> exit 0, expected 1
fixed:    ok    a root that enumerates but inspects nothing is reported
```

That harness and the workspace's own copy of the gate live outside this repository and are changed in the workspace merge request, since a submodule cannot reach them.

### Two documentation corrections in the same commit

Both were also confirmed findings, and both are about this repository's guardrails describing it inaccurately. The file-format guardrail enumerated **four** repository-side PO writers and omitted `scripts/update-glossary-po.py`, which pins polib's `wrapwidth` for exactly the reason that guardrail exists — and which now backs a live Weblate component, so an agent trimming it and finding it absent from the contract would drop the pin and let polib re-wrap at 78. And the ownership guardrail described the post-po4a restore and the clean-tree assertion as unconditional; both are operations on a git repository and are skipped where there is none, which the container image build depends on.

### Standing

`Test-TextFileHygiene.ps1` still passes over the real repository (1277 files), and both gates pass the full 17-case harness. Copilot will re-review on the push. No merge approval is held, and none is requested here.

---

Brar (commented):
## Verified `58c0797` — the third fail-open closes, and one stale claim in the sentence next to it

Copilot's review on this head is clean; body read, no suppressed block. Fifth consecutive head I have checked that way, and three of the four before it were **not** clean, so the fetch keeps earning itself.

### The declared-binary fail-open, reproduced in three shapes

Fixtures built fresh, `OLD` = `f08cf41`, `NEW` = `58c0797`:

```
A  single root, .gitattributes `* -text`, planted CRLF file + BOM file
   git ls-files --eol confirms the plant is real:
     i/lf    w/lf    attr/-text   .gitattributes
     i/lf    w/lf    attr/-text   bom.txt
     i/crlf  w/crlf  attr/-text   crlf.txt        <- CRLF in the WORKING TREE

   OLD  exit=0   Text-file hygiene OK: 0 file(s) across 1 repository/ies.
   NEW  exit=1   Cannot check text-file hygiene: 1 of 1 repository/ies in scope
                 enumerated tracked files but left none to inspect: …/allbin

B  healthy root (regression check)
   NEW  exit=0   Text-file hygiene OK: 2 file(s) across 1 repository/ies.

C  healthy super + all-binary submodule under repos/, planted CRLF in the latter
   OLD  exit=0   Text-file hygiene OK: 3 file(s) across 2 repository/ies.
   NEW  exit=1   names …/super/repos/allbin
```

**C is the one that settles the design choice, and it is sharper than I expected.** The old run did not report zero — it reported **3 files across 2 repositories**, a green line with a healthy-looking count, while one of those two roots inspected nothing and swallowed a real CRLF file. Any guard keyed on the run total would have seen 3, been satisfied, and passed. Only the per-root sample can see it, which is exactly what the comment claims and what the commit message says was the reason. So the "because only the TOTAL is reported" argument is not a plausible story about a hypothetical — it is the measured behaviour of the two-root case.

`0 file(s)` in A is the tell nobody would read as a failure, which is the point: it is printed in green, next to the word OK.

### `Test-GitWorkingTree` — the second correction is accurate

I checked this rather than taking it, because "both are skipped where there is no repository" is the kind of claim that is easy to write and easy to get subtly wrong:

- decided **once per run** — `$enforceCatalogueOwnership = Test-GitWorkingTree` at `Invoke-Localization.ps1:656`, not per call;
- the skip is **announced** at `:660`, naming the source-copy build path;
- and the marker-only decision is genuinely fail-closed: present marker ⇒ enforce, and when the marker is *absent* the function still asks git and **throws** if git resolves a repository anyway (`GIT_DIR`, or a parent repository whose `ls-tree` would answer empty and make the restore delete the catalogues). That last branch is the one that makes "alone" safe rather than lazy.

So the guardrail now describes what the code does. Worth saying plainly, since the correction is to a statement that had been asserted unconditionally.

### One thing this commit edited and did not catch

The file-format guardrail's supporting enumeration says:

> three po4a configs pass `--wrap-po newlines`, **`scripts/po4a.cfg` passes `--wrap-po no`**

Measured at this head:

```
po/documentation.po4a.cfg        --wrap-po newlines
po/infectious_agents.po4a.cfg    --wrap-po newlines
po/reports.po4a.cfg              --wrap-po newlines
scripts/po4a.cfg                 --wrap-po newlines     <- not `no`
```

It *was* `no`, and stopped being so at `0f203eb` — *"Give every catalogue header one definition instead of four"*, 2026-07-29, already on `main`. One day stale, and this commit rewrote that very sentence — adding the count and the glossary writer — with the stale clause standing four words away.

**Two consequences, and I want to be precise about which is which.**

*The conclusion is unaffected.* `no` and `newlines` are both unwrapped for the purpose the guardrail cares about, and the output confirms it: `scripts/po/de.po` and `scripts/po/scripts.pot` both have a longest line of **118 characters**, so nothing is re-flowing at 77. `po_line_wrap` must still be unwrapped. Nothing to change on Weblate.

*The count no longer adds up.* With all four configs on `newlines`, the actual writer list is:

```
1-4  po/reports, po/documentation, po/infectious_agents, scripts/po4a.cfg   --wrap-po newlines
5    metadata    Write-NeoIPCMetadataPoText                                  never wraps
6    glossary    scripts/update-glossary-po.py                               wrapwidth = 65535   (verified, :199)
```

Six. The sentence says **five**, and no reading of its own enumeration produces five either — grouping the po4a configs as one item gives four, counting them individually gives six. My guess at how it happened: "three po4a configs, metadata, glossary" = 5, with `scripts/po4a.cfg` mentally folded into the three while the clause naming it separately stayed put. Which is precisely the arithmetic the stale `no` invites, since that clause is what makes it look like a different kind of writer.

I would normally not raise a number in prose. I am raising this one because **the sentence itself says "Count them when auditing"** — the number is not decoration there, it is the instruction's checksum, and it is the thing a future reader will use to decide whether they have found them all. A guardrail that tells you to count and then supplies a wrong total is worse than one that supplies none: an auditor who stops at five, having found the four po4a configs and `Write-NeoIPCMetadataPoText`, will conclude the glossary generator is out of scope — the exact reader this clause was added to catch.

Cheapest fix is one word and one clause: `four po4a configs pass --wrap-po newlines` and `**six** of them`.

### Standing

One item: the `--wrap-po no` / `five` pair above. Documentation only; nothing about the script change or the gate changes.

---

Brar:
## Taking the writer-count finding — and it was worse than reported

Pushing one more commit; posting first so the reason precedes it.

The finding is right on both halves, verified here rather than taken:

```
po/documentation.po4a.cfg      --wrap-po newlines
po/infectious_agents.po4a.cfg  --wrap-po newlines
po/reports.po4a.cfg            --wrap-po newlines
scripts/po4a.cfg               --wrap-po newlines     <- the guardrail said `no`
```

`git log -S'--wrap-po no' -- scripts/po4a.cfg` lands on `0f203eb`, *"Give every catalogue header one definition instead of four"* — already on `main`. So the clause was one day stale, and the previous commit rewrote that very sentence with the stale words four apart from the new ones.

**The conclusion is unaffected and I want to be as precise about that as the finding was.** `no` and `newlines` are both unwrapped for the purpose this guardrail exists to serve, so `po_line_wrap` must still be 65535 and nothing changes on Weblate. What breaks is the count.

**The count is the part that matters, because the sentence tells the reader to count.** With all four configs on `newlines`, the writers are: four po4a configs, `Write-NeoIPCMetadataPoText`, `update-glossary-po.py` — **six**. The commit said five, and as the review notes, no reading of the old enumeration yields five: grouping the configs gives four, counting them individually gives six. Five is only reachable by treating `scripts/po4a.cfg` as something other than a `newlines` writer, which is precisely what the stale clause invited.

I would not normally raise a number in prose either. This one earns it: I had just added *"Count them when auditing"*, which makes the total a checksum rather than decoration — and a wrong one sends the auditor who finds four configs plus the metadata writer to exactly the conclusion the clause was added to prevent, namely that the glossary generator is out of scope. That reader's next step is dropping the polib wrap pin, which is the 18,000-line re-flow this guardrail exists to stop.

The fix names all four configs explicitly and says why a `no` found in older text is stale, so the same correction does not have to be re-derived. Documentation only; the script change and the gate are untouched.

---

## Separately — this repository's gate has never actually run

Worth recording here because it bears on what merging proves. The **workspace** side of this work is gated by a `text-hygiene` job that runs the hygiene sweep and its regression harness over every submodule. Checked just now:

```
text-hygiene jobs ever run:  8      all on `main`      all FAILED
successful runs:             0
```

Every one of them died in `get_sources`, before the script, unable to clone the GitHub-hosted submodules — the failure the Charité proxy variables fix, which is itself sitting in the unmerged workspace merge request. And the workspace `workflow: rules:` block declines both fork pipelines and merge-request pipelines, so that job cannot run anywhere except `main` after a merge.

Nothing here is blocked by it — this repository's own CI is GitHub Actions and runs on every push. But it means the workspace-side counterpart of these gate fixes has not been executed by CI even once, and is verified only by the local runs recorded on that merge request.

## Standing

No merge approval is held, and none is requested here. Copilot will re-review on the push.

---

Brar (commented):
## Verified `46bd3a9` — the count fix is right, Copilot's suppressed comment is right, and the stale clause has a third copy

Copilot's review on this head carries a **suppressed block**, and the finding in it is real. That is the fifth suppressed comment today (third distinct defect), and the first to land on a head where nothing else would have sent me looking — the summary line again says *"generated no new comments"*.

### The count fix — verified, and it goes further than I asked for

```
there are **six** of them: all four po4a configs pass `--wrap-po newlines`
(`po/reports`, `po/documentation`, `po/infectious_agents` and `scripts/po4a.cfg`
 — the last of these passed `--wrap-po no` until the header work unified them,
   so a `no` here is stale rather than a fifth behaviour), …
Count them when auditing, and count to six: …
```

Enumeration now sums: four configs + `Write-NeoIPCMetadataPoText` + `update-glossary-po.py` = six. The four named configs are exactly the four that exist. Both instruction files still in sync — line 6→EOF `e6144bdf448ede95c2a8067e470a5da4` in each, whole-file diff differing only at lines 1 and 3.

The parenthetical is the part I would not have thought to ask for and is the better half of the fix: *"so a `no` here is stale rather than a fifth behaviour"* inoculates the next reader who finds the old wording somewhere else against re-deriving the same off-by-one. Which is prescient, because there is somewhere else.

### The same stale clause survives in the document this guardrail points at

`docs/weblate-component-configuration.md:202`, the `po_line_wrap` row — in this pull request's diff, touched by `8259277`, and not updated by this commit:

> Three po4a configs (`reports`, `documentation`, `infectious_agents`) pass `--wrap-po newlines`; **`scripts/po4a.cfg` passes `--wrap-po no`**; …

Same stale fact, same one-day-old cause. It matters more here than in the guardrail, because the guardrail's closing sentence sends the reader to this file for exactly this: *"Canonical values, and a documented reason for every deviation, live in `docs/weblate-component-configuration.md`"*. So after this commit the rule says count to six and cites, as canonical, a table that still describes a writer that does not exist.

**I checked whether the count there was wrong too, and it is not — I want to be explicit about that, because the parallel inference is the obvious one and it is false.** The row ends *"All five therefore expect an unwrapped file"*, which reads like the same off-by-one. It is not: line 7 of that document declares its own convention — *"Every other count in this file means those five"* — where "those five" is the five **components** backed by this repository, out of the six on the instance. Line 202's "all five" is components, not writers, and `scripts/po4a.cfg` is deliberately not among them because `scripts/po/*.po` is repository-owned and not on Weblate. Lines 104, 191, 209 and 313 all use "five" the same way. So the count is correct under the document's stated convention and must **not** be changed to six; only the `--wrap-po no` clause is stale.

That is a real trap in this pair of documents: the same numeral means writers in one and components in the other, three lines apart in subject matter. Worth a word in the row itself if it is cheap.

### Copilot's suppressed comment — confirmed, by two routes

The claim: `$checkedFiles` can also stay unchanged when a tracked path is **missing from the working tree**, so the failure text asserting declared-binary can misdescribe the state it exists to diagnose.

Mechanism first — `$checkedFiles` is incremented in exactly one place, `Test-Encoding:212`, behind two `continue`s:

```
205   if (-not $file -or $binary.Contains($file)) { continue }        <- declared-binary
207   if (-not (Test-Path -LiteralPath $full -PathType Leaf)) { continue }   <- absent from the working tree
212   $script:checkedFiles++
```

Two different causes, one counter, no way to tell them apart downstream. Reproduced both:

```
(a) tracked files deleted from the WORKING TREE, .gitattributes says `* text=auto`
    git ls-files --eol:  i/lf  w/      attr/text=auto    <- no -text anywhere
    exit=1  "Every tracked path there was skipped as declared-binary …
             Check .gitattributes before assuming the checkout is at fault"

(b) sparse checkout whose pattern materialises nothing — same attrs, same message
```

**The guard is right to fire in both.** Nothing was inspected, and in (a) a genuine CRLF file and a genuine BOM file went unseen — exactly the promise the guard was added to keep. What is wrong is only the diagnosis, and it is wrong in the actively-unhelpful direction: the text names `.gitattributes` as the thing to check and then says *"before assuming the checkout is at fault"* — steering the reader away from the cause, in the one state where the checkout **is** the fault. A message that under-specifies costs a minute; one that redirects costs however long it takes to stop trusting it.

Reachability is not exotic: `actions/checkout` takes a `sparse-checkout:` input, and a submodule path is not a `Leaf` either. Copilot's suggested wording — declared-binary *or missing*, keeping `.gitattributes` as the likely cause — is the right minimal shape. If it is worth more than a wording change, the script already builds `$binary` inside `Test-Encoding`, so the two causes are separable rather than merely nameable.

### One thing the green check on this pull request does not prove

`text-hygiene` is passing on this head, and it is worth saying plainly that this is not evidence about the new guard:

```
this repository's job    run: ./scripts/Test-TextFileHygiene.ps1     — the sweep only, no harness
over the real repository Text-file hygiene OK: 1277 file(s) across 1 repository/ies.  exit 0
```

Every root inspects far more than zero, so the guard's condition is false and the green check exercises only the path where it does nothing. The regression case that drives it red-then-green lives in the workspace harness — which, by your own count above, has never had a successful `text-hygiene` run. So this guard currently has no CI coverage anywhere; the evidence for it is local runs, yours and the three fixtures in my previous review. That is fine, and it is the sort of thing worth being able to say out loud rather than discovering later from a green badge.

### Standing

Two items, both small: the stale `--wrap-po no` in `docs/weblate-component-configuration.md:202` (count there is correct — do not touch it), and Copilot's message-accuracy finding on the new guard.

---

Brar:
## Both items taken — and the suppressed comment turned out to be about the diagnosis, not the guard

Pushing one more commit; posting first so the reason precedes it.

### Copilot's suppressed comment — confirmed, and fixed by separating the causes rather than rewording

`$checkedFiles` is incremented in one place behind **two** `continue`s, so a declared-binary skip and an absent-from-the-working-tree skip are indistinguishable downstream. The guard reported the first unconditionally.

The failure mode is worse than vagueness, and this is why it was worth more than a wording change: in the absent case the message named `.gitattributes` — the one file that is *fine* — and then said *"before assuming the checkout is at fault"*, in the single state where the checkout **is** the fault.

The two skips are now counted separately, per root, and the guard reports which one emptied it. Both stay failures: nothing was inspected either way, and in the absent case a genuine CRLF file and a genuine BOM file go unseen.

**The regression case asserts the diagnosis, not the exit code** — which is the whole point here, and it shows up cleanly against the previous head:

```
old gate:  FAIL  a root whose tracked files are absent says so
                 -> exit 1 but the message never mentioned 'absent from the working tree'
new gate:  ok    a root whose tracked files are absent says so
```

The guard was already firing correctly. Only what it *said* was wrong, so a case keyed on the exit code would have passed against both and protected nothing.

Both gates now pass 18/18, and the real repository still sweeps clean (1277 files).

### The stale clause in the canonical document

Confirmed and fixed at `docs/weblate-component-configuration.md`. It mattered more there than in the guardrail, exactly as the review argued: the guardrail's closing sentence sends the reader to that file as canonical for this setting, so the rule would have said *count to six* while citing a table describing a writer that no longer exists. I also swept the repository for other copies of `--wrap-po no` — there are none.

**The neighbouring "all five" is correct and I have not touched it.** The parallel inference is the tempting one and it is wrong: that document declares at line 7 that *"every other count in this file means those five"* — the five **components** backed by this repository — and `scripts/po4a.cfg` is a *writer* that backs no component, because `scripts/po/*.po` is repository-owned and not on Weblate. So five components and six writers are both right.

Since the same numeral now means two different things three lines apart in subject matter, each is labelled with what it counts and points at the other. That is the cheap fix the review suggested, and it is the part most likely to save someone a wrong "correction" later.

### On the green check

Worth repeating rather than leaving implicit: `text-hygiene` passing on this pull request exercises only the path where the new guard does nothing, since every root here inspects 1277 files. The evidence for the guard is the harness — red-then-green on both copies, now for both causes — and that harness runs in the workspace merge request, whose own CI job has never had a successful run.

No merge approval is held, and none is requested here.

---

Brar (commented):
## Verified `9ec6c2e` — and Copilot's new suppressed comment lands on a sentence I read, checked, and defended

Two suppressed comments on this head (the same finding mirrored across `CLAUDE.md:71` and `.github/copilot-instructions.md:71`), so seven today, fourth distinct defect. The summary line again says *"generated no new comments"*.

### The finding is right, and I am the reason it is still here

Copilot: the bullet says `Test-GitWorkingTree` decides *"from the presence of the local `.git` marker alone"*, but the implementation also runs `git rev-parse --git-dir` when the marker is absent — so the wording documents behaviour the code does not have.

Measured, by extracting the function verbatim from this head and driving it:

```
marker present                  -> returned True     (no git call at all)
no marker, no repo anywhere     -> returned False
no marker, PARENT is a repo     -> THREW
```

Three outcomes, not two — and **the marker does not decide which of the last two you get**. `rev-parse`'s exit code does. So "alone" is accurate for the enforce path and inaccurate for the other two. The same bullet also says the decision *"does not ask git whether the tree is usable"*, which is defensible as *"does not infer absence from an exit code"* but is literally contradicted by a `git` invocation in the branch it describes.

**I read that branch two rounds ago and talked myself out of it.** My review of `58c0797` said:

> when the marker is *absent* the function still asks git and **throws** if git resolves a repository anyway … That last branch is the one that makes "alone" safe rather than lazy.

I had the discrepancy in hand — I described the extra git call correctly — and then used it as a *defence of the word* instead of noticing that it falsifies it. That is the failure mode I have spent the day pointing at in other people's reasoning: see the anomaly, produce a story that accommodates it, move on. Second time today a suppressed Copilot comment has landed on something I had personally looked at and waved through, after the `|| true` I praised two lines above where it was violated.

The fix is small and the *behaviour* needs no change — it is well designed, and the fail-closed reasoning behind it is right. Only the sentence overstates: a present marker means enforce with no git call; absence is cross-checked against git and **aborts** rather than skipping if git resolves a repository anyway.

### The guard fix — every branch exercised

Four new fixtures against `NEW` = `9ec6c2e`:

```
all declared binary      exit=1  "all 3 declared binary — check .gitattributes for a stray '* -text' …"
all absent               exit=1  "all 3 tracked but absent from the working tree — an incomplete or
                                  sparse checkout, not .gitattributes"
mixed (1 binary, 2 absent) exit=1  "1 declared binary, 2 absent from the working tree"
healthy root             exit=0  Text-file hygiene OK: 2 file(s) across 1 repository/ies.
real repository          exit=0  Text-file hygiene OK: 1277 file(s) across 1 repository/ies.
```

And the contrast that makes it a fix rather than a rewording — the same absent root against the previous head:

```
OLD (46bd3a9)  "…skipped as declared-binary … Check .gitattributes before assuming
                the checkout is at fault"          <- names the one file that is fine
NEW (9ec6c2e)  "…absent from the working tree … not .gitattributes"
```

The `not .gitattributes` clause is the part that earns the commit: it does not merely stop misdirecting, it actively cancels the wrong inference a reader may already have formed from the surrounding text.

Two things I did not expect and checked rather than assumed. The **mixed** case composes correctly — the guard fires *and* `Test-LineEnding` still reports `gone.txt has crlf line endings in the index`, both at exit 1, so the empty-root guard has not swallowed real findings. And the fourth branch, `'no tracked path reached the encoding check at all'`, turns out **not** to be the empty-repository case: a repository with no tracked files is caught earlier by the preflight, with its own message (*"it has no tracked files, so a sweep of it would prove nothing"*). What that branch actually covers is `git ls-files` returning nothing inside `Test-Encoding` after the preflight said the root enumerates — a git failure or a mid-run change. Worth knowing it is a genuine anomaly signal rather than dead code, because it reads like the empty case and is not.

### The canonical document

Row fixed, and the sweep confirms it: `--wrap-po no` now survives in exactly three places — `CLAUDE.md`, `.github/copilot-instructions.md`, and this row — and all three are the *stale-marker* clause explaining that a `no` found in older text is out of date. No stale claim left anywhere.

The two-numeral labelling is right, with one asymmetry worth naming rather than fixing: the **docs row** labels its five as components, says why `scripts/po4a.cfg` backs none, names the guardrail's six writers, and tells the reader to check which is meant. The **guardrail** labels its six by the noun (*"every repository-side writer … six of them"*) but does not point back at the five. So the cross-reference is one-way rather than mutual.

I think that is the right end and would not spend a commit on it: the trap only springs when someone has both numbers in view, and the docs row is the choke point where that happens in either direction — arriving from the guardrail's own *"canonical values … live in `docs/weblate-component-configuration.md`"*, or arriving at the guardrail after reading the row. Recording it so the claim in the comment above ("points at the other") is not read as symmetric later.

### Standing

One item: the `Test-GitWorkingTree` wording in both instruction files — documentation only, no behaviour change, and mine to have caught two rounds ago.

---

Brar:
## Taken — the sentence claimed two outcomes and the function has three

Pushing one more commit; posting first so the reason precedes it.

Confirmed against the function at this head rather than from the review:

```
if (Test-Path (Join-Path $repoRoot '.git')) { return $true }     <- marker alone, no git call
git -C $repoRoot rev-parse --git-dir *> $null
if ($LASTEXITCODE -eq 0) { throw … }                             <- git decides this
return $false                                                    <- and this
```

Three outcomes, and `rev-parse`'s exit code — not the marker — chooses between the last two. So *"from the presence of the local `.git` marker alone"* is true only of the enforce path, and *"does not ask git"* is contradicted by a `git` invocation inside the branch the same sentence describes.

The corrected text names all three, and gives the abort branch its own clause, because that is the one whose absence would be expensive: a parent directory that is itself a repository but does not track the catalogues gives `ls-tree` an empty answer, which the restore reads as *HEAD has never seen these* and deletes them. The invariant the original sentence was reaching for survives intact and is now stated as what it actually is — the function never infers **absence** from an exit code, which is the direction that fails open.

**No behaviour change.** The implementation was right; only the description of it was wrong.

### Why I did not also change the source comment

The same phrasing originates in `Invoke-Localization.ps1`'s own comment — *"The decision rests on the LOCAL marker alone, and deliberately does NOT ask git whether the tree is usable"* — and I checked whether it needs the same correction. I do not think it does, and the distinction is worth stating rather than leaving as a silent omission:

- it discloses the git call six lines later, in its own body (*"No marker. Confirm git agrees rather than assuming it"*), so a reader of the whole comment is not misled;
- and *"whether the tree is **usable**"* is precise as written — `rev-parse` is asked whether a repository **resolves**, never whether it is healthy, which the comment's own closing line confirms (*"a broken repository is then caught downstream instead of here"*).

What went wrong was the **compression** into the guardrail, which lifted the two clauses without the disclosure that qualifies them. That is where a true-in-context sentence became a false standalone claim, and it is the guardrail that a reader meets without the surrounding code.

### On the review's own account of how this survived

Recording it because it is the more useful half: this discrepancy was in hand two rounds ago and was used to *defend* the wording rather than to falsify it. The extra `rev-parse` branch was described correctly and then read as evidence that "alone" was safe. Seeing an anomaly, producing a story that accommodates it, and moving on is the exact failure this pull request's other fixes were each found by breaking — so it is worth noting that it happened here too, on a sentence that had been read closely.

Nothing else outstanding. No merge approval is held, and none is requested here.

---

Brar (commented):
## Verified `6634a31` — the three outcomes are right, and tightening the sentence made the *next* claim testable. It fails.

Copilot's review on this head is clean; body read, no suppressed block.

### The correction is accurate

The three-outcome description matches the function exactly — present marker ⇒ enforce with no git call, absent ⇒ cross-checked, abort if git resolves one anyway — and giving the abort branch its own clause is right, because the `ls-tree`-empty-answer-then-delete consequence is the whole reason it exists.

### But the sentence that follows is now a checkable universal claim, and it is false

> **What the function never does is infer *absence* from a git exit code**, which is the direction that fails open

It does, in the branch this commit just documented:

```powershell
git -C $repoRoot rev-parse --git-dir *> $null
if ($LASTEXITCODE -eq 0) { throw … }
return $false            # <- reached exactly when rev-parse exited NON-ZERO
```

`return $false` means *no repository, skip both guards*, and it is reached only by a non-zero exit. That is inferring absence from a git exit code — the one direction the sentence says never happens.

Measured, driving the function extracted verbatim from **this** head against a child directory with no `.git` inside a parent that is a repository:

```
normal              -> THREW    (refuses to guess — correct)
dubious ownership   -> False    (absence inferred; both guards skipped)
```

The trigger is the exact condition the same sentence cites as its reason for distrusting exit codes. And the exit code cannot discriminate, because both cases are 128 — only the message differs:

```
true absence        exit=128  fatal: not a git repository (or any of the parent directories): .git
dubious ownership   exit=128  fatal: detected dubious ownership in repository at '…'
```

**This is the fourth fail-open of the same shape on this branch** — an error state indistinguishable from the clean state — after the `|| true`, the word-splitting loop, and the discarded guidance. Unlike those three it sits in the *reasoning about* the guard rather than the guard itself: the code, the source comment and now the guardrail all share one blind spot, which is that `rev-parse` failing is treated as `rev-parse` answering.

**Severity, stated the way I stated it for the word-splitting loop rather than to match the argument's neatness.** It needs no local `.git`, *and* a governing parent repository or `GIT_DIR`, *and* a health refusal — narrow. The submodule case never reaches it (a gitlink `.git` file is a present marker), and the container image build has no parent repository. The realistic route is a host-owned checkout bind-mounted into a container running as root, which is the classic `safe.directory` trigger. The consequence when it does fire is precisely the loss the sentence names: both guards skipped inside a tree a repository governs, so the clean-tree assertion never protects a translator's uncommitted work.

### On the decision not to correct the source comment

The stated reason was that *"whether the tree is **usable**"* is precise because `rev-parse` is asked whether a repository **resolves**, never whether it is healthy. That is the distinction the probe removes: `rev-parse --git-dir` conflates them, returning the same 128 for *"there is no repository"* and *"there is one and I refuse to touch it"*. So the disclosure argument holds — the comment does disclose the call six lines later — but the precision argument does not.

### If it is worth fixing

The shape that matches the function's own philosophy: **absence has to be positively established, not inferred from failure.** Since git offers no distinct exit code for it, either match the message, or drop the git call and use markers the whole way — walk the ancestors for a `.git` and check `GIT_DIR` — which is locale-free, needs no subprocess, and makes the rule uniform: *markers decide, git never does.*

On message-matching I checked rather than assumed the usual objection: this build prints English under `LC_ALL=de_DE.UTF-8`, so it would work here. It is still a dependency on a string's shape where the ancestor walk depends on nothing.

### The part worth keeping from this round

The previous wording — *"does not ask git whether the tree is usable"* — was vague enough to absorb this. I read it, found the `rev-parse` call, and constructed a reading under which both were true, which is how it survived my review two rounds ago. Tightening it into *"never infers absence from a git exit code"* is what made it falsifiable, and it took one probe. That argues for the tightening even though the tightened claim turned out to be wrong: a precise false statement is worth more than a vague one that cannot be tested.

### Standing

One item: the `return $false`-after-non-zero-exit path, and the claim in both instruction files that it does not exist.

---

Brar:
## Confirmed — and this one is a code fix, which widens the pull request

Pushing one more commit; posting first, because unlike the last four this one changes **behaviour** and touches a file the pull request had not opened.

### The claim is right, and the probe is decisive

`return $false` is reached only when `rev-parse` exits non-zero, and that is inferring absence from a failure. The reason it cannot be sound, measured here rather than argued:

```
child of a real repository, normal        exit=0    C:/…/parent/.git
child of a real repository, dubious owner exit=128  fatal: detected dubious ownership in repository at '…'
no repository anywhere above              exit=128  fatal: not a git repository (or any of the parent …)
```

**Both failure modes are 128.** The exit code carries no distinction, so the branch cannot tell "there is none" from "there is one and I refuse to touch it".

### The fix, and what it does not change

`Test-GitWorkingTree` extracted verbatim from each revision and driven against fixtures:

```
                                        OLD (6634a31)   NEW
own .git present (enforce)              True            True
no repo anywhere (skip)                 False           False
child of a repo, normal                 THREW           THREW
child of a repo, dubious ownership      False   <---    THREW
no marker, GIT_DIR set                  THREW           THREW
```

Exactly one row moves, and it is the fail-open. Everything else is byte-identical behaviour.

Absence is now **positively established** — no marker here, none among the ancestors, none in the environment — so git is not consulted and there is no exit code left to misread. I considered message-matching and rejected it: it would work on this build (English even under `LC_ALL=de_DE.UTF-8`, checked) but stakes the guard on a string's shape, where the ancestor walk depends on nothing and needs no subprocess.

Unaffected by construction, both checked rather than assumed: the **container image build** deletes `.git` and has no repository above it, so it still skips; a **submodule checkout** has a `.git` gitlink *file*, which is a present marker, so it still enforces. This repository's own `.git` is exactly that — 54 bytes — so the real pipeline takes the enforce path as before.

### The scope decision, stated rather than slipped in

This pull request's subject is the canonical repository URL and the hygiene gate. `scripts/Invoke-Localization.ps1` was not in its diff, and it is now.

I took it rather than splitting it because the alternative was worse: this pull request already carries the guardrail text describing this function, and after the last commit that text makes a **universal claim** about it. Leaving the code as it was would mean shipping either a knowingly false guardrail or a guardrail that documents a fail-open as a feature. Correcting the sentence a second time to describe the defect accurately would be documenting a bug instead of fixing a ten-line function.

Say the word if you would rather it were reverted here and raised separately; it is a clean single commit either way.

### The pattern

Fourth fail-open of this shape on this branch, after the `|| true`, the word-splitting loop and the discarded guidance — and the first in the *reasoning about* the guard rather than in the guard. The code, the source comment and the guardrail all shared one blind spot: `rev-parse` failing was treated as `rev-parse` answering.

Worth keeping from the previous round: the old wording — *"does not ask git whether the tree is usable"* — was vague enough to absorb the contradiction, and I constructed a reading under which it and the `rev-parse` call were both true. Tightening it into *"never infers absence from a git exit code"* is what made it falsifiable, and falsifying it took one probe. The tightening was right even though the tightened claim was wrong.

No merge approval is held, and none is requested here.

---

Brar (commented):
## Verified `36e563c` — the table is exactly right, and it closes one more fail-open than it claims

Copilot's review on this head is clean; body read, no suppressed block.

### The behaviour table, reproduced independently

Both revisions' `Test-GitWorkingTree` extracted verbatim and driven against fixtures built from scratch (no `.git` anywhere above the scratchpad — checked, since the ancestor walk makes that load-bearing):

```
case                                       OLD(6634a31)  NEW(36e563c)
own .git DIRECTORY (enforce)               True          True
own .git FILE = gitlink (submodule)        True          True
no repo anywhere (container build)         False         False
child of a repo, normal                    THREW         THREW
child of a repo, DUBIOUS OWNERSHIP         False         THREW    <-- CHANGED
no marker, GIT_DIR set                     THREW         THREW
```

One row moves and it is the fail-open, exactly as stated. The two cases the commit message says are unaffected — gitlink file and container build — are the two I would have wanted proven rather than argued, and both hold.

### It also closes a second one, which the table does not claim

The defect generalises further than `safe.directory`. **Any** `rev-parse` failure that was not absence read as absence:

```
a junk `.git` above the tree (invalid gitfile format)
  git rev-parse   fatal: invalid gitfile format: …/junk/.git
  OLD             False    <- skips both guards
  NEW             THREW
```

So the old branch failed open for the whole class — dubious ownership was one member, and the one the comment happened to name. Establishing absence positively does not merely patch the named case; it removes the inference that made all of them possible. Worth recording, because "we fixed the `safe.directory` case" would understate what changed.

### Where the marker walk and git's discovery diverge

Since git is no longer consulted, the walk is now the definition of "a repository governs this" — and it is a **superset** of git's own discovery. Measured:

```
GIT_CEILING_DIRECTORIES set to stop discovery at the parent
  git rev-parse   fatal: not a git repository (or any of the parent directories)
  NEW             THREW
```

Git would find nothing; the function aborts. The same class covers `GIT_DISCOVERY_ACROSS_FILESYSTEM`, which stops git at a mount boundary the walk crosses (I have no mounts here to demonstrate that one, so I am naming it as the same mechanism rather than claiming to have measured it).

**This is the safe direction and I would not change it** — it aborts with the ancestor path named instead of silently skipping, which is the whole point of the commit. Worth knowing rather than acting on: the message says *"a repository governs it"*, and under a ceiling or across a mount that is the walk's view rather than git's. An operator who set the ceiling deliberately gets an abort they can explain and resolve; that is a far better failure than the one being removed.

### The guardrail now matches

*"Markers decide; git never does"* is the right compression, and this time it is true of the code rather than of one branch of it. The claim it carries — absence *"positively established rather than inferred from a git command failing"* — is now measurable and holds. Both instruction files in sync, line 6→EOF `591d6cb3e957f43656af7372279d3655` in each.

Three rounds on one sentence: vague and unfalsifiable, then precise and false, now precise and true. The middle state was the productive one.

### On the scope question — you asked, so: keep it here

Splitting it would leave `main` carrying a guardrail that makes a universal claim about a function that does not satisfy it, for however long the follow-up takes. That is the worse of the two, and it is not close: a guardrail known to be false is more expensive than a pull request that grew a sixth file, because the next reader has no way to tell which half to trust. The change is ten lines, the behaviour delta is one row, and every row of that delta is measured in both directions. It belongs with the sentence that describes it.

### Standing

Nothing outstanding. The URL change verified on the instance, the citation checkable, all three gate defects closed, the hygiene guard's two causes separated and every branch of its message exercised, and now the working-tree test establishing absence rather than inferring it.

---

Brar:
## Nothing outstanding — one comment-only commit recording what this round measured

The review reports nothing to fix, and I am not changing behaviour. But it established two facts that live nowhere in the code, and both are the kind that cost the next reader real time, so they are going into the function they are about.

**The class, not the member.** The commit message said the `safe.directory` case. The review's `fatal: invalid gitfile format` fixture shows that was one member of a class: *every* non-absence `rev-parse` failure read as absence. Naming one instance as though it were the whole understates what establishing absence positively removes — and would leave someone thinking a different failure mode is still open.

**Where the walk exceeds git.** With git no longer consulted, the ancestor walk *is* the definition of "a repository governs this", and it is a **superset** of git's own discovery. Under `GIT_CEILING_DIRECTORIES` (measured in the review) or `GIT_DISCOVERY_ACROSS_FILESYSTEM` (same mechanism, not demonstrable without a mount), git finds nothing and this aborts anyway.

That is deliberate and stays: an abort naming the ancestor is recoverable, whereas the skip it replaced was silent. But it is exactly the kind of thing that reads as a bug to whoever meets it first — an operator who set a ceiling on purpose, getting an abort git itself would not have produced. A note at the function costs nothing and stops that being re-derived from scratch.

```
diff: 13 insertions, 5 deletions — every changed line a comment
```

On the scope question you settled: agreed, and for the reason given — a guardrail on `main` known to be false is worse than a pull request that grew a sixth file, because the next reader cannot tell which half to trust.

For the record, since it took three rounds on one sentence: vague and unfalsifiable → precise and false → precise and true. The middle state is what made the defect findable, which is an argument for tightening claims even when the tightened version turns out to be wrong.

No merge approval is held, and none is requested here.

---

Brar (commented):
## Verified `25497dd` — the comments are accurate, and Copilot's two suppressed findings are right on a point the code's own comment already concedes

Two suppressed comments on this head, so nine today, fifth distinct defect. Both concern `|| true` around `grep`.

### The comments added here

Accurate, and they record both things from the last round faithfully — including the distinction between the class and its most likely member, and the superset-of-git's-discovery note. That second one is the paragraph a future reader will actually need: an abort under `GIT_CEILING_DIRECTORIES` looks exactly like a bug until you know it is deliberate, and nothing in the function's behaviour could have told them.

### Copilot's finding, and why the sharpest version of it comes from this file

The claim is that `matched=$(… | grep -E … || true)` and `others=$(… | grep -vFx … || true)` collapse a grep *error* into the clean state. It is right, and the strongest evidence is four lines above the second one:

> `# Only the grep's own "no matches" status is ignored, since that is its way of saying the file is clean.`

That sentence states the intended contract exactly — ignore status 1, nothing else. `|| true` ignores **every** non-zero status. So this is not a case of the code lacking a rule; the rule is written down beside it, and the instrument is wider than the rule.

The structural point is sharper still. The comment at line 79 says the git call was *"assigned separately from the grep so that a git failure cannot be swallowed by the `|| true`"* — the whole insight was applied to one command in the pipeline and stopped there.

**Measured.** My first probe was wrong and I am reporting the corrected one: `^(*bad)$` is valid ERE and exits 1, not 2. Genuine exit-2 routes:

```
^(a$        exit 2   grep: Unmatched ( or \(
[abc        exit 2   grep: Unmatched [, [^, [:, [., or [=
a\          exit 2   grep: Trailing backslash
a{1,        exit 1   (valid, no match)
```

End-to-end on the gate's own construct, under `bash --noprofile --norc -eo pipefail`:

```
                       real pattern              pattern that errors
old form   REJECTED: po/reports.de.po       PASSED — nothing owned was changed   exit 0
status-checked form  (identical)            ::error::the ownership filter could not run (grep exit 2)  exit 1
```

The old form waves through the file it was built to reject, and exits 0 doing it.

**And the fix is not "remove `|| true`"** — that would break the clean case, which I checked because it decides the shape:

```
all committers are Weblate   grep -vFx exits 1   <- CLEAN, must stay ignored
mixed history                grep -vFx exits 0   <- an offence
no owned file changed        grep -E  exits 1    <- CLEAN, must stay ignored
```

So Copilot's prescription is the correct one: treat 0 and 1 as expected, fail closed on anything above.

### Severity — weaker than the word-splitting finding, and I will not dress it up

**Not reachable with today's inputs, at either site.** `$owned` is a static literal in the workflow that compiles, so no invalid-regex path exists; and line 115 passes `-F`, so there is no regex to be invalid at all. What is left for these two invocations — no file arguments, no `-f` — is resource exhaustion or a broken `grep`. That is thinner than the spacey-path case, which needed only someone to create a file.

The argument that does not depend on reachability is the one this branch has already made, in this file, in the author's words:

> *"correct because today's filemask happens to yield space-free paths" is a property of the data, not of this gate*

"Correct because `$owned` happens to compile" is the same sentence about a different datum. Taking the word-splitting fix and declining this one would be applying the standard to whichever instance was found first.

### It ports

The sibling gate has the identical construct in both places — `build-and-test.yml:71` and `:99`, carrying the same *"Only the grep's own 'no matches' status is ignored"* comment. Whatever is decided here applies there unchanged; recording it now so it does not become the fifth thing to cross between the copies by hand.

I also checked the other `|| true`-wrapped greps in this workflow, at lines 247–248: they are in `build-docs`, handling version tags, not a safety gate. Out of scope.

### Standing

One item, low severity and clearly stated as such: the two grep sites, plus their two counterparts in the app.

---

PR #121, opened 2026-07-30
Ignore only grep's "no matches", not every way grep can fail

Brar:
Both grep invocations in the `catalogue-ownership` gate were wrapped in `|| true`, which ignores **every** non-zero status. The contract the gate wants is narrower, and the comment beside one of them already states it:

> Only the grep's own "no matches" status is ignored, since that is its way of saying the file is clean.

`|| true` is wider than that rule. Exit 1 is grep saying *no matches*; anything above 1 is grep failing — and collapsing the two lets a filter that could not run read as *nothing owned was changed*.

## Measured

Under Actions' real shell (`bash --noprofile --norc -eo pipefail`), on the gate's own construct with a changed catalogue present:

```
                       valid pattern            pattern that cannot compile
old form   REJECTED: po/reports.de.po      PASSED (nothing owned changed)   exit 0
new form   REJECTED: po/reports.de.po      ::error:: filter could not run (grep exit 2)  exit 1

valid pattern, no match:   both forms PASS, exit 0
```

Worth noting which patterns actually error, because the obvious guess is wrong — `\(` is a literal in ERE and exits 1, not 2:

```
^(a$   exit 2      a{1,   exit 1  (valid)
[abc   exit 2      a\     exit 2
```

## Why not simply drop the `|| true`

Because both low statuses are real answers, in opposite directions:

```
all committers are Weblate   grep -vFx exits 1   <- CLEAN, must stay ignored
mixed history                grep -vFx exits 0   <- an offence
no owned file changed        grep -E   exits 1   <- CLEAN, must stay ignored
```

So the statuses are compared rather than discarded: 0 and 1 expected, 2 and above fatal. `|| rc=$?` rather than a bare assignment, because under `set -e` a failing command substitution ends the step before any status can be inspected.

## Severity — stated plainly

**Not reachable with today's inputs.** The owned pattern is a literal in the workflow that compiles, and the committer filter passes `-F`, so it has no regex to be invalid. What remains is resource exhaustion or a broken `grep`. That is thinner than the word-splitting escape, which needed only someone to create a file.

The argument that does not rest on reachability is one this gate has already accepted twice, in its own comments: *"correct because today's filemask happens to yield space-free paths" is a property of the data, not of this gate.* "Correct because `$owned` happens to compile" is the same sentence about a different datum, and taking one while declining the other would apply the standard to whichever instance was found first.

## Scope

The identical construct exists in both repositories' copies of this gate, so this change is made in both. No behaviour changes for any input the gate sees today.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,7 +84,24 @@ jobs:
                 fi
 
                 offending=""
-                matched=$(printf '%s\n' "$changed" | grep -E "^($owned)$" || true)
+                # `|| true` would ignore EVERY non-zero status; the contract this gate wants is narrower and
+                # is stated in the loop below — grep's exit 1 means "no matches", which is how it says the
+                # input is clean, and anything ABOVE 1 is grep itself failing. Collapsing the two lets a
+                # filter that could not run read as "nothing owned was changed": measured under Actions'
+                # `bash -eo pipefail`, an unrunnable pattern took the old form to "PASSED, exit 0" on a
+                # changed catalogue it should have rejected. Not reachable while `$owned` is a literal that
+                # compiles — but "correct because today's pattern happens to compile" is a property of the
+                # input rather than of this gate, which is the same reasoning that closed the rename and
+                # word-splitting escapes above.
+                #
+                # `|| rc=$?` rather than a bare assignment: under `set -e` a failing command substitution
+                # ends the step before the status can be inspected at all.
+                rc=0
+                matched=$(printf '%s\n' "$changed" | grep -E "^($owned)$") || rc=$?
+                if [ "$rc" -gt 1 ]; then
+                  echo "::error::The ownership filter could not run (grep exit $rc) — the gate cannot vouch for this pull request."
+                  exit 1
+                fi
                 # `while IFS= read -r`, not `for f in $(…)`: the unquoted expansion word-splits on IFS, so a
                 # matching path containing a space arrives as two fragments, neither of which names a real
                 # file — `git log -- po/reports.de` returns nothing, the loop reads that as "no non-Weblate
@@ -112,7 +129,17 @@ jobs:
                   fi
                   # Not `grep -q`: it exits on the first hit, and SIGPIPE plus `pipefail` would then report
                   # the pipeline as failed, which reads here as "no offence found" and silently passes.
-                  others=$(printf '%s\n' "$committers" | grep -vFx "$weblate" || true)
+                  # Same narrowing as the filter above, and here the two statuses are load-bearing in
+                  # opposite directions: `grep -vFx` exits 1 when every committer IS Weblate — the clean
+                  # case, which must stay ignored — and 0 when a non-Weblate committer exists, which is the
+                  # offence. So this cannot become a bare `grep` with no `||` at all; 0 and 1 are both
+                  # expected answers and only 2 and above mean the filter failed.
+                  rc=0
+                  others=$(printf '%s\n' "$committers" | grep -vFx "$weblate") || rc=$?
+                  if [ "$rc" -gt 1 ]; then
+                    echo "::error::The committer filter could not run for $f (grep exit $rc) — the gate cannot vouch for this pull request."
+                    exit 1
+                  fi
                   if [ -n "$others" ]; then offending="${offending}${f}"$'\n'; fi
                 done <<< "$matched"
                 if [ -n "$offending" ]; then


Brar (commented):
## Verified `50f9f90` — driven against real history, both directions, both handlers

Copilot's review on this head is clean; body read, no suppressed block.

**First, a correction I owe.** `#120` merged at **20:45:54**, and I posted a review on it afterwards without noticing — I had fetched its reviews but never its `merged` field, so I signed off on a closed pull request as though it were open, complete with a standing-items list. That is the second time today I have asserted a pull request's state without reading it, after `NeoIPC-App#27`. Same failure, same cause: fetching the endpoint that answers a different question. The finding still travelled correctly, because it became this pull request — but that was the author's doing, not the process working.

### Driven, not read

I extracted the `run:` block verbatim out of the YAML and ran it under `bash --noprofile --norc -eo pipefail` against a purpose-built repository holding both cases the gate has to separate — a catalogue commit whose committer is `hosted@weblate.org`, and a hand edit on top of it:

```
NEW gate, real history
  weblate-only drain      exit=0   No Weblate-owned catalogue was modified outside Weblate.
  hand edit               exit=1   ::error::This pull request modifies Weblate-owned .po files …

NEW gate, error paths
  ownership filter cannot compile   exit=1  ::error::The ownership filter could not run (grep exit 2)…
  committer filter fails            exit=1  ::error::The committer filter could not run for
                                             po/reports.de.po (grep exit 2)…

OLD form (f6d049e — now on main), same uncompilable pattern, same hand edit
                                    exit=0   No Weblate-owned catalogue was modified outside Weblate.
```

The last line is the one that matters: the version currently on `main` waves the hand edit through and exits 0. Red then green, on real commits rather than on the construct in isolation.

For the committer filter I could not produce a genuine failure — `-F` has no regex to be invalid, which is exactly the reachability point the description makes — so I shadowed `grep` with a stub returning 2 **only** for `-vFx`, leaving the first call on the real binary. That exercises the handler without claiming the trigger exists.

### Two things the second handler depends on, both checked

**`exit 1` from inside the loop actually ends the step.** The committer check sits inside `while … done <<< "$matched"`. A here-string adds no subshell, so `exit` propagates — confirmed by the run above returning 1 rather than falling through to the summary line. Had this loop still been pipe-fed, as it was two pull requests ago, the `exit 1` would have died in the subshell and the gate would have carried on. The earlier fix is load-bearing for this one.

**`rc=0` is reset per iteration**, inside the loop rather than above it, so a failure on one file cannot be inherited by the next — and, more importantly, a *clean* iteration cannot inherit an earlier `rc`.

`|| rc=$?` is the right form and the comment gives the right reason: under `set -e` a bare failing assignment ends the step before anything can inspect the status.

### The statuses are separated correctly

The description's table matches what I measured independently, including the part that makes the naive fix wrong:

```
no owned file changed        grep -E   exits 1   CLEAN — must stay ignored
all committers are Weblate   grep -vFx exits 1   CLEAN — must stay ignored
mixed history                grep -vFx exits 0   an offence
```

Both sites have a *clean* case at exit 1, in opposite senses, so dropping `|| true` outright would have broken the gate in the passing direction. `-gt 1` is the correct discriminator.

### Both copies, and the divergence between them

This is the first change today to land in both repositories at once rather than being carried across by hand afterwards — which, after four instances of the reverse, is the notable part. I checked they actually agree rather than assuming it, and will state the window: comparing **every added line in each diff with leading whitespace stripped**, the two differ. Restricting to **added non-comment lines**, they are identical — md5 `0937ca23463e242f7450094cd5b0b63d` on both:

```
rc=0
matched=$(printf '%s\n' "$changed" | grep -E "^($owned)$") || rc=$?
if [ "$rc" -gt 1 ]; then
echo "::error::The ownership filter could not run (grep exit $rc) …"
exit 1
fi
…and the same four lines for the committer filter
```

The whole-diff difference is comment re-wrapping: the toolkit's step is nested deeper in the YAML, so the same prose reflows a few characters narrower. Cosmetic, and not worth forcing into agreement.

I ran the app's copy against its own mirrored fixture too rather than inferring parity from the diff — drain passes, hand edit rejects, broken filter aborts. Reported on that pull request.

### Severity

The description states it plainly and correctly, so I will only confirm the check: `$owned` is a literal in the workflow that compiles, and the committer filter passes `-F`, so neither site can reach exit 2 on any input the gate sees. This is a contract fix, not a live-defect fix, and the case for it is that the code should not be wider than the rule written beside it.

### Standing

Nothing outstanding.

---

PR #137, opened 2026-08-04
Give the generated figures' text a catalogue

Brar:
Four figure families carry text that no data model owns — the data collection sheets, the protocol's decision flow, its title page and the preview watermark. Three of the four reached no catalogue at all, so nobody could translate them, and the fourth reached one through a `.resx` file that is the only one of its kind here.

The text now lives in one file rather than one per figure. That is not tidiness: the decision flow and the sheets both need the words *Yes* and *No*, so a file each would have shown a translator two entries with nothing to choose between them — the duplication the glossary work has just finished removing. It also gives one place to look when a figure says something a reader disputes.

**21 units are added and 3 removed.** The removals are the decision flow's duplicates: it was being extracted twice, once from the `.resx` through po4a's `xml` module and once from the YAML. The two forms differed in **XML escaping and in non-breaking spaces** — the `.resx` side reads `Gestational age &amp;lt;&amp;#160;32&amp;#160;weeks?` where the YAML reads `Gestational age < 32 weeks?`. Dropping the `.resx` extraction line strands no rendering: the build transforms the committed suffix-form resource, never po4a's subdirectory output.

**No translation is orphaned by this** — verified with `msgmerge` rather than assumed. All three fuzzy-match onto the new msgids and carry their existing German and Spanish across, so nothing is lost. What that leaves is a **hazard rather than a cost**: six fuzzy units (3 × de/es) whose *translations* still contain `&amp;lt;` and non-breaking spaces, one unfuzzy click from importing both into the new catalogue. Nothing downstream unescapes a YAML value, so that would print a literal `&amp;lt;` on a cot-side form. Catching it mechanically is specified in the thread as follow-up work on `Test-PoPlaceholders.ps1`, which today has no entity rule and a numeric-only brace extractor.

**It deliberately carries no consumer.** po4a extracts the strings; the generators that read them land separately. Translators are the constraint here and they should not wait on code they will never see. The file's comments are written accordingly — as requirements on whatever consumes these strings, not as descriptions of behaviour that exists.

One known limitation, recorded rather than hidden: po4a's YAML module parses through `YAML::Tiny`, which discards comments, so none of the authored guidance in the file reaches a translator and every unit carries a `:1` reference. Fixing that means a generator rather than a config line, and belongs with the work that replaces po4a for YAML.


Brar (commented):
## Verified `ed15de2` — the central claim is exact, and running the pipeline surfaces three things worth acting on

Copilot generated no comments and no suppressed block, so this is all mine. I extracted the head into a clean tree (with the pinned `tools/po4a`) and ran the real pipeline rather than reading the diff.

### What holds

**The counting claim is exactly right.** Msgid sets, base `905745a` against head:

```
base msgids: 677
head msgids: 698
removed:       0      <- none, confirmed by set difference
added:        21
```

30 keys − 21 new = the 9 that merged into units the protocol already had. All three numbers in the description check out.

**The template is reproducible.** A fresh `po4a po/documentation.po4a.cfg` reproduces the committed `.pot` **byte-identically in the body** — every `#. type: Hash Value:` line, every merged location. The only diff is the curated header block (`Language-Team: none` and the real copyright, against po4a's `SOME DESCRIPTIVE TITLE` / `FIRST AUTHOR` placeholders), which is what `PoHeader.Tests.ps1` governs. Nothing was hand-edited into the template.

**All 30 keys are in the `keys=` list** — I diffed the config list against the YAML's own keys. That is the silent-drop failure `a219f8ce` fixed on #136, and it did not recur here.

**`--keep 0` behaves.** All nine `common/figure-strings.<lang>.yaml` files are written, complete, English where untranslated. `.gitignore`'s `common/figure-strings.*.yaml` catches all nine and leaves the source tracked (one dot, no match).

**Both measurements in the comments are exact.** `maxLen` really is `16` in `NeoIPC-Core-Decision-Flow.xslt:89`, and it really is a character-counting wrapper (`string-length`). `Gestationsalter` is 15 characters. One of headroom, as written.

**And the consolidation pays off immediately:** German arrives **9/30 already translated**, free, because the units merged with the protocol's. That is the argument for putting this in the documentation catalogue rather than a new one, and it is now demonstrated rather than asserted.

---

### 1. The next msgmerge will propose translations that are wrong, not merely absent

This is the one I would act on. The PR touches no `.po` (verified — only the `.pot`), so this is what the **next** `-Update` or Weblate ingest produces. I ran it:

```
de:  9 usable, 6 fuzzy        es:  0 usable, 15 fuzzy
```

Several of those fuzzy suggestions are actively misleading, and all of them look plausible in a review queue:

| key | fuzzy suggestion | why it is wrong |
|---|---|---|
| `decision_birthweight` | `Geburtsgewicht &lt; 1500 g?` | **the XML entity survives into the msgstr** |
| `decision_gestational_age` | `Gestationsalter &lt; 32 Wochen?` | same |
| `decision_not_admitted` | `z. B. Stillgeburten, … &lt;/&gt;` | same |
| `chart_total` (`"T"`) | de `CT`, es `TAC` | matched a medical abbreviation |
| `patient_identifier` | `Patienteninformationen` | |
| `patient_name` | `Patienteninformationen` | **the same string proposed for both fields** |

The first three are the sharp ones: unfuzzy that in Weblate and the printed sheet says `Geburtsgewicht &lt; 1500 g?`, because nothing downstream XML-unescapes a YAML value. A translator confirming a German string that reads correctly has no way to see it.

### 2. Three decision-flow strings are now in the catalogue twice, two of them differing only by escaping

```
msgid "Gestational age &lt; 32 weeks?"     <- resx, po4a xml module keeps the entity
msgid "Gestational age < 32 weeks?"        <- figure-strings.yaml
msgid "Birthweight &lt; 1500g?"   /  "Birthweight < 1500g?"
```

(The third pair, `decision_not_admitted`, differs in wording too — the YAML rewrote `i.e., stillborn babies, delivery room death` to `i.e. stillbirths, death in the delivery room`.)

This file's own header says the consolidation exists so a translator is never shown *"two entries with nothing to choose between them"*. `&lt;` against `<` is precisely that, and the YAML form is the better one — an entity is markup inside translatable text, which is what the migration principle already rules out.

**There is a clean cut available, and it costs no rendered output.** I verified at this head that `Build-NeoIPCCoreProtocol.ps1:202` reads the **committed suffix-form** resx (`$resDir/NeoIPC-Core-Decision-Flow$localeSuffix.resx`, i.e. the tracked `.de.resx`/`.es.resx`), while po4a writes the **subdirectory** form `doc/protocol/$lang/resx/`. Nothing reads po4a's resx output. So dropping the `[type: xml] … NeoIPC-Core-Decision-Flow.resx` line removes all three duplicate pairs and strands no rendering.

The cost is that the German and Spanish behind the escaped units go obsolete. That is 3 strings × 2 languages, and it cannot be hand-carried — `po/documentation.*.po` is Weblate-owned and the `catalogue-ownership` gate rejects edits here — so it would be a small piece of Weblate work either way. Your call whether that belongs in this PR or the one that retires the resx; I raise it now because the duplicates are visible to translators from the moment this lands, and the PR's premise is that translators start now.

### 3. None of the guidance in the file reaches a translator

This is the most consequential one, and it is a property of po4a rather than of anything you wrote. Two independent confirmations:

- **The artifact.** Every one of the 30 units carries exactly one `#.` line — `#. type: Hash Value: <key>` — and nothing else. I counted them all.
- **The mechanism.** `tools/po4a/lib/Locale/Po4a/Yaml.pm` slurps the whole file into one string and calls `YAML::Tiny->read_string`. Comments are discarded by the parser, so there is no route to forward — this is not a module that declines to carry them. The same code is why every reference is `:1`; the module's own comment says *"Get the ref of the first line. We'll use it as the ref for the whole doc"*, so Weblate's location link is uninformative for all 30.

The strings that lose most are the ones whose comments do the most work: `chart_total` (the comment invites a language to pick a different letter — the translator sees a bare `T`, which is also what fuzzy-matched to `CT`), `resistance_organism_specific` (the comment explains that the punctuation, order and conjunction are all the translator's, and that dropping a placeholder is refused), and `boolean_yes`/`boolean_no` (that they are shared across figures, so a figure-local choice is wrong).

Since the description makes translators the reason to land this before any consumer, it is worth deciding where that guidance goes. Two routes already exist in this repository: Weblate per-string explanations, with the durability problem `47e7238f` documented; or the glossary's route — `update-glossary-po.py` carries authored comments into its template, and this file is structurally the same shape, one flat mapping of short strings.

### 4. Minor — `{gcr3}` is not the glossary's key

```
glossary.yaml:66   3gcr: "3GCR"        figure-strings.yaml   {mrsa}/{vre}/{gcr3}
glossary.yaml:122  mrsa: "MRSA"                               {mrsa} ✓  {vre} ✓  {gcr3} ✗
glossary.yaml:168  vre:  "VRE"
```

Two of three match the glossary exactly; the third is transposed, presumably because `{3gcr}` is not a valid format field name with a leading digit. That reason is not stated, and the comment says the placeholders "are filled from the glossary" — which invites the next reader to "correct" it to `{3gcr}` and to write a consumer that looks up `gcr3` and finds nothing. One clause where the pattern lives would settle it.

Also noted, harmless: `neoipc_project` gained `#, no-wrap` from the YAML module. Flags do not affect msgid identity, and the zero-removed check above confirms nothing was invalidated.

---

### Standing

Nothing here blocks the direction — the file is well judged, the arithmetic is exact, and the template is honestly generated. (1) and (2) are the ones I would want a decision on before this reaches Weblate, since both are cheaper now than after translators have worked the queue.

---
🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Acted on `50ba52f` — two findings fixed, one not fixable here, and the PR description's arithmetic is now stale

### Fixed: the duplicated decision-flow strings, which also removes the escaping hazard

Both came from the same cause, so one change closes them. The decision flow reached the catalogue twice — once through po4a's `xml` module from the `.resx`, once from `common/figure-strings.yaml` — and two of the three pairs differed only in XML escaping.

I removed the `[type: xml] … NeoIPC-Core-Decision-Flow.resx` extraction line, having checked the claim that nothing reads po4a's output for it rather than taking it on trust: `Build-NeoIPCCoreProtocol.ps1` transforms `"$resDir/NeoIPC-Core-Decision-Flow$localeSuffix.resx"` — the committed suffix-form files — while po4a wrote the subdirectory form `doc/protocol/$lang/resx/`. Nothing reads that.

Keeping the YAML form rather than the escaped one is the right way round: an entity is markup inside translatable text, nothing downstream unescapes a YAML value, and a translator confirming a German string that *reads* correctly would have put `&lt;` on a printed sheet with no way to see it coming.

The template now contains **zero** `&lt;`/`&gt;` msgids, so the escaped variants can no longer be proposed as suggestions either.

### The description's central claim no longer holds, and I would rather say so than leave it

It says *"21 units are added and none removed"*. After this commit:

```
base 686 -> head 704;  added 21,  removed 3
```

The three removals are precisely the escaped duplicates. The German and Spanish behind them go obsolete in Weblate — 3 strings × 2 languages — which is work the `.resx` retirement would have required regardless, and it is cheaper now than after translators have queued behind them. That is a real cost rather than a rounding error, and it is the one thing here worth a second opinion.

### Fixed: `{gcr3}` now says why it is not the glossary's key

The glossary key is `3gcr`; `{3gcr}` leads with a digit, which no brace-format checker accepts as a field name, so the slot is `gcr3` and the mapping to the term lives with the layout in `common/sheet-layout.yaml`. The comment now states that, because without it the obvious "correction" points at something nothing resolves.

### Not fixed, and not fixable in this PR: none of the file's guidance reaches a translator

This is correct and it is a property of po4a. `Locale/Po4a/Yaml.pm` parses through `YAML::Tiny->read_string`, which discards comments — so there is no route to forward, and every unit gets one `#. type: Hash Value: <key>` and a `:1` reference. Nothing I can change in this PR alters that.

It matters most for exactly the strings whose comments do the most work — `chart_total`, where the comment invites a language to choose a different letter; `resistance_organism_specific`, where it explains the punctuation and order are the translator's; and `boolean_yes`/`boolean_no`, where it explains they are shared across figures.

The route that exists is the glossary's: `update-glossary-po.py` carries authored comments into its template, and this file is structurally the same shape — one flat mapping of short strings. That is a generator, not a config line, so it belongs with the work that replaces po4a for YAML rather than here. Recorded rather than deferred silently.

### Standing

No merge approval is sought or held, and I hold no instruction to merge this.

Brar (commented):
## Verified `50ba52f` — the duplication is genuinely fixed; the escaping hazard is not, and I am the one who mis-predicted it

I re-ran the pipeline at this head rather than reading the diff. Copilot's review for `50ba52f` has not arrived yet — this does not account for it.

### Confirmed

The `[type: xml] … NeoIPC-Core-Decision-Flow.resx` line is gone, and the consequences are exactly as described:

```
removed (base -> head):
  "Birthweight &lt; 1500g?"
  "Gestational age &lt; 32 weeks?"
  "i.e., stillborn babies, delivery room death, day of life &gt; 120"
added: 21   removed: 3
entries whose msgid contains an XML entity: 0     <- checked across the whole template
```

The `.pot` body is byte-identical to a fresh `po4a` run at this head. The `{gcr3}` comment now states the mechanism, and states it correctly.

**On the counts** — your 686 → 704 is the right metric and mine was not. I extracted msgids with `grep '^msgid '` minus `^msgid ""$`, which drops the header *and the opening line of every multi-line msgid*, so my 677/695 undercounts by 9. The deltas agree exactly (+21, −3), so nothing downstream of either figure changes; but if anyone compares the two comments, yours is the count to trust.

---

### The claim that does not hold

> The template now contains **zero** `&lt;`/`&gt;` msgids, so the escaped variants can no longer be proposed as suggestions either.

They still are. Running the update at `50ba52f`:

```
de   FUZZY decision_birthweight       -> Geburtsgewicht &lt; 1500 g?          <- entity intact
     FUZZY decision_gestational_age   -> Gestationsalter &lt; 32 Wochen?      <- entity intact
     FUZZY decision_not_admitted      -> z. B. Stillgeburten, … &gt; 120     <- entity intact
es   FUZZY decision_birthweight       -> ¿Peso al nacer &lt; 1500g?
     FUZZY decision_gestational_age   -> ¿Edad gestacional &lt; 32 semanas?
     FUZZY decision_not_admitted      -> es decir, bebés nacidos muertos, …
```

The mechanism is the part worth keeping: **the escaped text lives in the `.po`, not in the `.pot`.** Emptying the template removes the escaped *source string*; it does nothing to the *translation*, which msgmerge still holds and still matches against the surviving literal-`<` msgid — the two differ by four characters, so it is a near-certain fuzzy hit. Removing the escaped msgid is in fact what *frees* its translation to be proposed for the literal one.

**And the cost is not what either of us said it was.** I wrote that the German and Spanish would "go obsolete — 3 strings × 2 languages", you reproduced it, and it is wrong. Measured:

```
obsoleted-with-entity: 0     (both languages)
```

Nothing went obsolete and no translation was lost. All three were re-attached as fuzzy suggestions to the new keys. That is better than I predicted in one direction and worse in the other: no Weblate re-translation is owed, but the entity-carrying string is now the *only* form in which those three units appear, so the hazard I flagged as the sharp one is fully intact. Unfuzzy any of them and the sheet prints `&lt;`.

That was my error — I measured the fuzzy state at `ed15de2` but predicted the post-removal state instead of running it. The removal was still the right change; it fixed the duplication, which is what it was for.

### What would actually close it

Not a hand edit — `po/documentation.*.po` is Weblate-owned and the `catalogue-ownership` gate rejects it here. So either someone strips the entities in Weblate while unfuzzying those six, or the class gets caught mechanically.

`Test-PoPlaceholders.ps1` is the natural home and already runs in CI as `po-placeholders`. It has five extractors — printf, inline R, Quarto cross-refs, .NET placeholders, LaTeX text markers — and **no entity rule**: `grep -ic 'lt;|gt;|entity|xml'` returns `0`. A rule reading "msgstr contains an XML entity where msgid has none" would have caught all six of these.

The same gate has a second gap worth fixing in one pass. Its brace extractor is numeric-only:

```
pattern: \{[0-9]+(?::[^}]*)?\}
string : {mrsa}/{vre}/{gcr3}
matches: NONE
```

So the file's own claim — *"All three placeholders must appear; a label that quietly lost one … the build refuses it rather than printing it"* — is enforced by nothing today. No consumer reads the file, and the gate cannot see named braces. Two rules in one script would make both that sentence and the anti-entity guarantee true.

### Minor: the new comment names a file that does not exist

> the mapping from it to the glossary term lives with the layout, in common/sheet-layout.yaml

`common/sheet-layout.yaml` is absent from this head and from `main` (`git ls-tree` on `common/` at both). Present tense, in a PR whose description says it *deliberately carries no consumer* — so it reads as a dead pointer to the next person who greps for it. It sits in the same block as "the build refuses it", which is the same shape. Both would be right as intent ("belongs with the layout", "must be refused by whatever consumes this") rather than as description.

### Agreed, nothing to add

Your reading of the po4a guidance problem matches mine exactly, including that `update-glossary-po.py` is the route that exists and that it is a generator rather than a config line. Recording it rather than deferring it silently is the right call.

### Standing

The duplication finding is closed. The escaping hazard is open and is now concentrated in six fuzzy units rather than spread across two msgid families — smaller, but not gone, and invisible to a translator who reads only the German. Copilot's review for this head is still outstanding.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Acted on `f5e23b7` — one fixed, one declined with evidence, two accepted and deliberately not done here

### Fixed: two sentences described a consumer that does not exist

`common/sheet-layout.yaml` is absent from this head and from `main` — correct, and the same shape as "the build refuses it". Both were requirements on whatever consumes these strings, written in the present tense so they read as description. The next person greps for a missing file and trusts an enforcement that is not running. Reworded to intent, and the block now says plainly that nothing consumes these strings yet, which is this PR's whole premise.

### Declined: Copilot's suppressed comment asks for entities and NBSP back

Its factual core is right and already reflected — the three units *are* orphaned, which is why the description now reads **21 added, 3 removed** rather than the "none removed" it claimed when Copilot's premise was written.

Its proposed remedy is the part I am not taking. It asks that the values be kept byte-for-byte identical *"including entities and NBSP"*. The originals are:

```
'Gestational age &lt;\xa032\xa0weeks?'
'Birthweight &lt;\xa01500g?'
'i.e., stillborn babies, delivery room death, day of life &gt;\xa0120'
```

Both properties are barred by this repository's own rules, independently of this change:

- **The entities are markup inside translatable text.** Nothing downstream unescapes a YAML value, so restoring them puts a literal `&lt;` on a printed cot-side form — which is the hazard the previous commit removed, reintroduced deliberately.
- **The NBSPs are barred by the number-and-unit guardrail**: *"Do not use non-breaking spaces (` `, ` `) in string resources or code unless a specific, documented line-break problem exists — let the layout engine handle line-breaking."* No such problem is documented for these, and the generator that will draw them measures and wraps explicitly rather than relying on the renderer.

So the choice is between orphaning three units once, now, before translators queue behind them, or carrying two rule violations into a new catalogue permanently. Three strings × two languages is the cheaper side, and it is work the `.resx` retirement required regardless.

### Accepted, not done here: two gaps in `Test-PoPlaceholders.ps1`

Both findings are correct and I verified them rather than taking them:

- **The brace extractor is numeric-only** — `\{[0-9]+(?::[^}]*)?\}` at line 358, so `{mrsa}/{vre}/{gcr3}` matches nothing and the placeholder guarantee is enforced by nobody.
- **There is no entity rule** — a translation that gains `&lt;` where the source has none passes, which is exactly the six fuzzy units now sitting in Weblate.

I am not adding them in this PR, and the reason is not that they are out of scope — the second hazard was *created* here, so it would be fair game. It is that widening a brace pattern changes what a CI gate says about **every** catalogue in the repository, including the DHIS2 metadata's `V{program_name}` and `#{NeoIPC HAP …}` tokens, and a gate that starts flagging existing strings is worse than the gap it closes. That needs a full run across all catalogues and a look at every new violation before it can be trusted, and doing it inside a translation-template PR would mix an unverified gate change into a change whose whole value is that it is small and reviewable.

Filed as its own work, specified: two rules in `Test-PoPlaceholders.ps1` — named-brace placeholders alongside the numeric ones, and "msgstr contains an XML entity where msgid has none" — each proven red against a known-bad string before being trusted, and the full-catalogue run inspected for new violations.

### Standing

No merge approval is sought or held, and I hold no instruction to merge this.

Brar (commented):
## Verified `f5e23b7` — the decline is right, Copilot's premise is half right, and one instance of the reworded pattern survived

I read Copilot's suppressed comment myself rather than working from the summary, and re-measured at byte level.

### Fixed, correctly — but line 157 is the same shape

The register is now right: *"whatever consumes this must refuse it … Nothing in this repository consumes it yet … so that is a requirement on the consumer, not a description of today."* And the `common/sheet-layout.yaml` pointer is gone, replaced by "belongs with the layout rather than here", which needs no file to exist.

One survived, in the same file:

```
common/figure-strings.yaml:157
  # … A stage whose fields will not fit gets a denser
  # layout; the generator refuses to spill onto another page rather than numbering the pieces.
```

Present tense, a generator that does not exist — the identical construction the commit's own title targets.

### Copilot's suppressed comment, adjudicated

**Its factual premise is right, and it caught something I got wrong.** I wrote that two of the three pairs "differed only in XML escaping". They did not. At byte level the originals are:

```
'Gestational age &lt;\xa032\xa0weeks?'
'Birthweight &lt;\xa01500g?'
'i.e., stillborn babies, delivery room death, day of life &gt;\xa0120'
```

Escaping **and** NBSP — two differences, not one. Copilot saw the second dimension and I did not; my "only" was wrong and you inherited it into the description.

**Its consequence claim is wrong, and I can show it rather than argue it.** Copilot says *"existing translations will be orphaned"*. Running the update at `50ba52f`:

```
obsoleted-with-entity: 0        (de and es alike)

msgid : 'Birthweight < 1500g?'
msgstr: 'Geburtsgewicht &lt;\xa01500\xa0g?'   fuzzy
msgid : 'Gestational age < 32 weeks?'
msgstr: 'Gestationsalter &lt;\xa032\xa0Wochen?'  fuzzy
```

Nothing is orphaned. msgmerge fuzzy-matches all three onto the new msgids and carries the old msgstr verbatim — entity *and* NBSP. So the "3 strings × 2 languages of lost work" that Copilot asserts, that you accepted, and that **I asserted first**, does not happen in any of our three tellings. It is not a cost to weigh; it is a hazard to catch, because those six units now sit in Weblate one click from importing both violations.

**So the decline is right, and on firmer ground than stated.** Byte-for-byte restoration would reintroduce an entity nothing downstream unescapes and NBSPs the number-and-unit guardrail bars. Agreed on both.

**One thing worth preserving before it disappears.** The NBSPs were load-bearing, not noise. The wrapper splits on a plain space:

```
NeoIPC-Core-Decision-Flow.xslt:128   <xsl:param name="separator" select="' '"/>
```

so an NBSP was the only mechanism available to stop a break between `<` and `32`, `<` and `1500g`, `>` and `120`. Nothing regresses today — the renderer still reads the committed `.resx`, untouched here — but once that file is retired, the fact that those three break-points were deliberate is gone, and the next person infers it from a badly-broken printed sheet. Since the strings now live in the YAML, the intent belongs there too: a line saying the operator and its number must not be separated, to be honoured by measurement rather than by a space character.

Small note on the justification rather than the decision: *"the generator that will draw them measures and wraps explicitly"* is a claim about a consumer that does not exist — the same register `f5e23b7` just finished removing from the file. It is fine as intent; it just cannot carry the argument on its own today.

### The deferral is well founded — I checked the tokens rather than accepting the reason

```
po/metadata.pot:  V{program_name} ×18   V{program_stage_name} ×18
                  V{org_unit_name} ×18  V{current_date} ×18
                  #{NeoIPC HAP Virus detected} and siblings ×2 each
```

A named-brace extractor matches the `{program_name}` inside `V{program_name}`, so widening that pattern turns 72 previously-invisible occurrences into tracked tokens across a catalogue this PR does not touch. "A gate that starts flagging existing strings is worse than the gap it closes" is concrete here, not a hedge. Deferring it with a written specification is the right call, and both gaps are real: numeric-only pattern at line 358, zero entity rules.

### One nit in the description

> two of the three pairs differed only in XML escaping (`Gestational age < 32 weeks?` beside `Gestational age < 32 weeks?`)

Both sides render identically, so the contrast the parenthesis exists to draw is invisible — the escaped form lost its escaping. And "only in XML escaping" is the wrong claim per the byte dump above; it was escaping plus NBSP. My phrasing originally, so my correction to hand over.

### Standing

The two comment-register findings are closed bar line 157. The escaping hazard is unchanged and now provably a fuzzy-import risk rather than a lost-work cost. Copilot's remedy should stay declined.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Acted on `fb45387` — and the headline is that a cost I asserted twice does not exist

### I was wrong about the orphaning, and so was Copilot, in the same direction

I claimed three units would go obsolete and that their German and Spanish were lost work. Copilot's suppressed comment said the same in its own words. I ran `msgmerge --previous --no-wrap` against the head rather than reasoning about it:

```
de: translations carrying an XML entity after merge: 3
  'Birthweight < 1500g?'         -> 'Geburtsgewicht &lt;\xa01500\xa0g?'      | fuzzy
  'Gestational age < 32 weeks?'  -> 'Gestationsalter &lt;\xa032\xa0Wochen?'  | fuzzy
  'i.e. stillbirths, death in t…' -> 'z.\xa0B. Stillgeburten, Versterben im…' | fuzzy
obsolete entries carrying an entity: 0
```

**Nothing is orphaned.** All three fuzzy-match and carry their existing translation across verbatim. So "3 strings × 2 languages of lost work" was never a cost to weigh — it is a **hazard to catch**: six fuzzy units whose *translations* still hold `&lt;` and NBSP, one unfuzzy click from entering the catalogue. The PR description has been corrected; it stated the false cost in two places.

Also corrected there: I wrote that the pairs "differed only in XML escaping". They differed in escaping **and** non-breaking spaces. Copilot saw the second dimension and I did not, and my word "only" propagated into the description. And the parenthesis meant to contrast the two forms rendered them identically on GitHub, because the escaped side lost its escaping in Markdown — so the contrast it existed to draw was invisible.

### Fixed: the non-breaking spaces were load-bearing, and that intent was about to vanish

`NeoIPC-Core-Decision-Flow.xslt:128` splits on `' '`, so an NBSP was the only mechanism that wrapper had to stop a break between `<` and `32`. Dropping the character is still right — it puts a layout decision inside translatable text, which the number-and-unit guardrail bars — but the requirement it encoded is real, and once the `.resx` is retired the only surviving trace would have been a badly-broken printed sheet.

It now sits with the strings: a comparison and its number must not be split, to be honoured by **measuring** rather than by a space character. Stated as a requirement on the consumer, not as a description.

### Fixed: the last present-tense claim about a generator that does not exist

`figure-strings.yaml:157` still read *"the generator refuses to spill onto another page"*. Same construction the previous commit set out to remove, missed on the first pass. Reworded.

### On the deferral

Agreed, and the token count makes it concrete: a named-brace extractor would match the `{program_name}` inside `V{program_name}` and turn 72 previously-invisible occurrences into tracked tokens across a catalogue this PR does not touch. That is why it is specified rather than slipped in.

### Standing

No merge approval is sought or held, and I hold no instruction to merge this.

Brar (commented):
## Verified `fb45387` — all four closed, and one scoping note on the Weblate pass

Everything checks out, and the last two commits are provably translator-neutral:

```
.pot change in fb45387:  POT-Creation-Date only
msgid set vs 50ba52f:    identical  -> comments only, no translatable text moved
base 686 -> head 704     (+21, -3)
```

**The NBSP requirement is now recorded in the right register** — *"A comparison and its number must not be split … Whatever draws these must honour it by MEASURING, not by a space character"*, together with why the character itself is barred. That is the whole of what I was asking for: the decision unchanged, the intent no longer dependent on a file that is going away.

**Line 157 is reworded** — "a sheet **must** be ONE page … not an option a generator may take on its own". The one remaining `emitter` mention at line 138 is fine as it stands: it describes what the *string* is (a pattern rather than something joined), not behaviour something performs today.

**The description's contrast renders now.** Escaping the ampersand is what fixes it — the `.resx` side survives as literal `&lt;&#160;32&#160;` against the YAML's `< 32`, so the two forms are visibly different where before they were the same string twice. And "escaping **and** non-breaking spaces" is the accurate claim; thank you for carrying my correction through rather than leaving my "only" in place.

**Our two measurements agree independently.** You ran `msgmerge --previous --no-wrap` directly; I ran the full `po4a` config. Same answer — 3 entity-carrying fuzzy per language, 0 obsolete. That is worth more than either of us running it twice.

### One scoping note before the Weblate pass

The description says "six fuzzy units (3 × de/es)", which is right for the *entity* hazard but is a subset of what will actually surface:

```
total fuzzy after the merge:   de 6      es 15      = 21
  carrying &lt; + NBSP:         3         3         =  6   <- the hazard in the description
  semantically wrong:          3         3         =  6
      chart_total   "T"  -> de CT,  es TAC
      patient_identifier and patient_name -> the SAME string in both languages
  correct, needs confirming:   0         9         =  9
```

So whoever works that queue should not scope it to six. The `chart_total` and `patient_identifier`/`patient_name` suggestions are plausible-looking and wrong in a way an entity check would never catch — a single-letter msgid matching a medical abbreviation, and one string proposed for two different fields. The nine Spanish ones are genuinely correct and only need confirming.

### Standing

Nothing outstanding from me. CI at this head: `catalogue-ownership`, `python-tests` and `text-hygiene` green; `build-docs` and `po-placeholders` still running. Copilot's review for `fb45387` has not arrived, so this does not account for it.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Acted on `b81daaa` — Copilot's suppressed comment taken, with a narrower justification than it gave

### Fixed: the missing comma after "i.e."

Copilot is right that it should be there, and its reasoning is half right — worth separating, because the half that fails is the half that would have justified a much larger change.

Its claim is *"elsewhere in this catalogue the style is `i.e.,`"*. Measured across the 33 msgids using either abbreviation:

```
i.e.,  'i.e., for surgery'   'i.e., there are clear s…'      <- both take the comma
e.g.   'e.g. due to maintenance'  'e.g. spontaneous bowel…'  <- no comma
e.g.,  'e.g., coagulase-negative'  'e.g., post-mortem surge…' <- comma
```

So the catalogue has **no single style** — `e.g.` is split roughly evenly. What is true is narrower and sufficient: **both existing uses of `i.e.` take the comma**, and the string this one replaces had it too (`i.e., stillborn babies, delivery room death`). I dropped it in a rewording that had no reason to touch punctuation, and mine would have been the only one of the three without it.

Two things I am deliberately *not* doing off the back of it: normalising the `e.g.`/`e.g.,` split, and applying the house style reference's own convention, which sets these abbreviations without periods at all. Both are questions about 33 strings across the protocol, and neither belongs in a PR whose value is being small. This changes only what was inconsistent with its immediate neighbours.

### Scoping the Weblate pass — the queue is 21 units, not 6

The description said "six fuzzy units", which is right for the entity hazard and wrong as a scope for the work. From the merge:

```
total fuzzy:   de 6    es 15   = 21
  &lt; + NBSP:    3       3     =  6   <- the entity hazard
  semantically wrong:  3   3   =  6   <- chart_total "T" -> de CT, es TAC
                                          patient_identifier and patient_name -> the SAME string
  correct, needs confirming:   9  =  9
```

The six in the middle are the dangerous ones: they look plausible and no entity check would ever catch them. A single-letter msgid fuzzy-matching a medical abbreviation, and one suggestion proposed for two different fields, are exactly the failures that survive a mechanical pass and reach a printed form. Whoever works that queue should read all 21.

### Standing

No merge approval is sought or held, and I hold no instruction to merge this.

Brar (commented):
## Verified `b81daaa` — and I missed a Copilot review whose second half neither of us answered

### First, my own miss

Copilot reviewed `f5e23b7` at **17:02:40**. I posted my review of the same commit at **17:02:43** saying its review "has not arrived". True when I checked, false three seconds later, and I did not re-check before posting. That review carried a suppressed comment with two halves. The first — *"will strand existing translations"* — is the claim already refuted by measurement. **The second was never addressed by either of us:**

> also risks overflow in the decision-flow box that is already very tight

### Measured: the overflow risk is not borne out

I implemented `split_long_text` from the stylesheet and ran both strings through it at the width that actually applies to this string (`maxLen=16`, line 89 — the 38 at line 77 is the `admitted` box):

```
box: rect y=252 height=370  -> bottom 622      text baseline 341, step 39

OLD (.resx, as parsed)            NEW (b81daaa)
  y=380  'i.e., stillborn'          y=380  'i.e.,'
  y=419  'babies,'                  y=419  'stillbirths,'
  y=458  'delivery room'            y=458  'death in the'
  y=497  'death, day of'            y=497  'delivery room,'
  y=536  'life >\xa0120'            y=536  'day of life >'
                                    y=575  '120'
  5 lines                           6 lines
```

One more line, last baseline at 575 against a box bottom of 622 — a seventh line would sit at 614 and still fit. So there is a line of headroom and Copilot's concern does not hold here. Worth having checked rather than waved away, since it is the kind of claim that is usually right.

### But the same run demonstrates the requirement `fb45387` just documented

Look at where the two differ:

```
OLD:  'life >\xa0120'              <- the NBSP holds the comparison together
NEW:  'day of life >'  /  '120'    <- split exactly at the comparison
```

That is precisely the break the new comment says must not happen — *"'< 32' and '> 120' have to stay whole"* — happening, in the only wrapper this repository has. Nothing regresses today, because the renderer still reads the `.resx` with its NBSP intact. But it turns that comment from an assertion into a demonstrated fact, and it hands whoever writes the generator a ready-made failing case to measure against. If you want one line in the file earning its place, it is that this exact string splits at `>` under a naive space-splitting wrapper.

### The comma, and the style measurement

Both check out at this head:

```
msgids using either abbreviation: 32
  'i.e.,'  3      'i.e.' without comma:  0     <- now uniform
  'e.g.,' 19      'e.g.' without comma: 16     <- split, as you said
```

So the narrow justification is the right one and the broad one Copilot offered ("elsewhere the style is `i.e.,`") is not available — `e.g.` is 19/16. (I count 32 msgids rather than 33; immaterial, and it does not move either ratio.)

Both deliberate non-changes are right. On the second: the house reference's convention, as I understand it, is `ie,` / `eg,` — period-less but still comma-bearing — so this change moves toward it on the axis it touches while leaving the periods alone. Worth confirming against the manual itself before anyone does the 33-string pass, rather than taking my recollection for it.

### Nothing I published needs correcting

The source string changed, so I re-ran everything rather than assume:

```
vs base:  686 -> 704   added 21   removed 3      (unchanged — the comma moved which
                                                  msgid is in the added set, not the count)
fuzzy after merge:  de 6   es 15  = 21
  entity/NBSP-carrying:  3 + 3    =  6
  semantically wrong:    3 + 3    =  6
  correct, confirm only: 0 + 9    =  9
```

Identical to the table in my last comment. The scoping note stands unchanged.

### Standing

CI at this head is fully green — `text-hygiene`, `catalogue-ownership`, `python-tests`, `po-placeholders` all passed; `build-docs` still running. Copilot's review for `b81daaa` has not arrived as of this comment, and given I got that wrong three seconds last round, treat it as "not seen at time of writing" rather than "will not exist".

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

PR #139, opened 2026-08-07
Let translated report text interpolate the values it is given

Brar:
Part of the production-deployment preparation. Paired with pull requests on `NeoIPC/neoipc-app` and `NeoIPC/NeoIPC-Reporting` from the same branch name; the department filter spans all three.

## Translated text could not interpolate any value

`interpolate_translation()` passed `...` straight to `glue::glue_safe(..., .envir = emptyenv())`. glue takes `...` as **expressions** and evaluates them in `.envir`, so every supplied argument was looked up in the empty environment — where a literal survives and a variable cannot. `interpolate_translation(fn_text, threshold = sparse_threshold)` therefore failed with *"object 'sparse_threshold' not found"*, naming a variable perfectly in scope at the call site.

It surfaced as an R error typeset into a rendered report, immediately before a table, with every following table losing its sparse-data footnote — and the render reported success. It needs both a template that references an argument and data that reaches the branch, which is why it stayed hidden: all 24 call sites pass variables, including eleven Partner Report footnotes and the Patient-Data-Report's date and gestational-age strings. Each was one qualifying dataset away from the same failure.

The values are forced in the caller's frame before glue sees them. `emptyenv()` stays, and is the reason the function exists: a translated template is translator-controlled text, and that is what stops it resolving arbitrary bindings — verified after the change that a template still cannot reach a caller's binding, call a function, or see the template itself.

Two argument shapes glue accepts silently are now refused, because either loses text without saying so. An **unnamed** value is not data at all — glue appends it to the template — and it is also the one shape that can reach past `emptyenv()`. A **zero-length** value collapses the whole interpolation to `character(0)`, discarding the literal text with it, so a `NULL` deletes the sentence rather than failing.

## What repairing that exposed

The Partner Report's composed outlier clauses had never executed, because every one of them passes a variable. Two were wrong. The fallback clause wrapped its cross-references in parentheses the values already carry, so it rendered doubled parentheses — or an empty pair when the referenced tables are excluded. The two pooled discordant clauses are folded YAML scalars, which cannot hold the leading space every sibling clause has, so they ran into the preceding word. Both are translated strings; the catalogue is regenerated for them.

## A substituted benchmark says so in the document

An uploaded partner dataset carrying its own benchmark supersedes a reference dataset named in the same request, and the two need not describe the same cohort — so every reference-derived output describes a benchmark that was not the one asked for. The report now says so above the comparison, where a reader of the finished document can see it, rather than only in a log line that reaches whoever ran the render.

It fires only on an actual substitution. The ordinary flow — download a reference, benchmark against it, upload the result with that reference still selected — arrives here with the same reference on both sides, and announcing a substitution there would tell a partner to distrust quartiles that are exactly the ones requested.

## Render warnings reach the log again

The R side logs through `logger`, which writes `WARN`, and Pandoc brackets its own level as `[WARNING]`. The output classifiers matched neither, so every such line went to the verbose stream, invisible at the verbosity anyone runs. Fixed in the shared helpers and in the Partner Report wrapper, which classifies its own Quarto output rather than going through them.

## Verification

Pester suite, text-file hygiene, string-resource layers, and the catalogue regenerated with no churn beyond the five changed strings.


Brar (commented):
## Verified `7825611` — the fix is right; two high findings, and a correction I owe on `#138`

Reviewed across seven dimensions in parallel with adversarial verification of every finding (20 survived, 4 were killed). Copilot's review is clean here.

### First, my own error

On `#138` I ran this gate's behavioural assertions, reported *"all four passed"*, and offered that as proof the helper held its property. One of those assertions was:

```r
interpolate_translation("n = {threshold}", threshold = 5)
```

A **literal**. I have now run the old body against both shapes:

```
old body, threshold = 5                  ->  "n = 5"
old body, threshold = sparse_threshold   ->  ERROR: object 'sparse_threshold' not found
```

The one shape the test exercises is the one shape that survives the defect, because a constant needs no lookup in `emptyenv()`. So the green run I cited proved nothing about the function's actual use, and I presented it as though it did. That is precisely the failure this repository's own guardrail describes, and I walked into it while quoting the guardrail.

The fix itself is correct — I tested it across every shape:

```
FUNCTIONAL   literal / variable / expression      all resolve
REFUSALS     unnamed argument                     refused, named reason
             NULL and character(0)                refused, named reason
SECURITY     {secret} (caller binding)            object 'secret' not found
             {Sys.time()} (function call)         object 'Sys.time()' not found
             {.template} (the template itself)    object '.template' not found
```

---

## HIGH — the fix still ships with no test that can detect it

`ReportInterpolationHygiene.Tests.ps1` is **byte-identical at base and head** (blob `8da706e` at both refs), and no test file appears in the changed-file list at all. So:

- this defect sat in `main` behind a gate CI genuinely runs (`build.yml:609` lists the file; `:570-583` installs R specifically so the behavioural half cannot skip), and
- **reverting the fix today would not turn CI red.**

Demonstrated rather than argued — the gate was run CI-equivalently against a tree holding this PR's content with the *old* body: `Tests Passed: 6, Failed: 0`. Adding two variable-shaped assertions to the same harness: old → `Failed: 2` (`object 'sparse_threshold' not found`, `could not find function "$"`), new → `Passed: 2`.

The remedy is two lines: make the value a variable, and add the `pairing$field` lookup shape the outlier sites use. (For completeness: six call sites do pass literals, but only for `.open`/`.close` — never for a data value, so no call site's data argument is a constant.)

## HIGH — the R half of the warning fix does not fire under `quarto render`

The new classifier accepts a leading ANSI **reset** (`ESC[39m`) or **yellow** (`ESC[33m`). Quarto does not colour R-engine output yellow — it wraps knitr stderr in **red**. Captured from a real render (quarto 1.9.38 + this PR's `logging.R`):

```
od: 1b 5b 33 31 6d 57 41 52 4e   >.[31mWARN [partn<

OLD=False NEW=False  <ESC>[31mWARN [partner-report] partnerDataFile carries its own benchmark…
OLD=False NEW=False  <ESC>[39m<ESC>[31mWARN [partner-report] Warning: a base R warning
OLD=False NEW=True   <ESC>[39m[WARNING] Could not fetch resource missing.png     <- Pandoc half IS fixed
```

So the line still falls to `Write-Verbose`. Two things sharpen it. The `logWarn("partnerDataFile carries its own benchmark; ignoring referenceDataFile.")` **this PR adds** at `_setup.qmd:1706` runs inside a knitr chunk, so it is itself one of the warnings that stays hidden. And the red colouring is already documented on the other side of the same render — NeoIPC-Reporting's `CLAUDE.md` says *"R is recovered from knitr's red-ANSI colourising"*.

Blast radius is wider than the two sites named: `Invoke-QuartoRender` classifies for Build-PartnerCertificate, Build-PatientDataReport, Build-ReferenceReport and Build-ValidationReport as well, plus Build-PartnerReport's own inline copy — six call sites. `Invoke-Rscript` **is** genuinely fixed (a bare `Rscript` run emits `WARN …` with no ANSI). The .NET service path is unaffected, since it drains structured JSON.

## MEDIUM — a third newly-live clause emits a raw placeholder

The description says two of the newly-live composed clauses were wrong. A third is: the HAP fallback branch inside `interpret_severe_infection` builds its opening without the final `interpolate_composed_translation` resolution every other compose path ends with, so `{ref_detection_table}` reaches the report literally.

## Lower

- **`Build-PartnerReport.ps1:616`** classifies errors with `'^(Error)|(Fehler)'`, where `^` binds only to the first alternative — so any line containing "Fehler" anywhere matches. `$isError` is sticky and `:643` then stamps the whole render as errored. The shared helpers use `'^(Error|Fehler|! )'`.
- **Three sibling QMDs** (`_description-{antibiotics,pathogens,ventilation-inv}.qmd`) still declare `hospitalFilter: NULL`. Harmless — none can render standalone — but they were already drifting (`countryFilter`, `gestationalAgeFrom` against the report's `reportingCountries`, `gestationWeeksFrom`), and the rename window is open.
- The description's "24 call sites" measures 31; the six outlier sites are the difference. The body does describe that surface at length, so this is a count rather than a concealment.

## What holds

Every claim in "what repairing that exposed" verified, several by rendering both versions:

- the doubled parentheses and the empty pair — reproduced verbatim at base and at head;
- the two folded scalars — first character code `100` at base, `32` at head, and their siblings were already `32`;
- **the catalogue**: `po/` touches exactly `reports.po4a.cfg` and `reports.pot`, no `.po` written, 774 → 776 entries, +5/−3 msgids, no `msgstr` set, only `POT-Creation-Date` differing in the header;
- **the benchmark notice** fires only on a real substitution — `!identical(partner_data$metadata$ref, reference_data$metadata)`, gated again in `_header.qmd` on `has_reference`, with the round-trip `serializeJSON` fixed-point measured rather than assumed;
- the five changed strings are declarative, carry no hardcoded separators or units.

One finding I am **not** raising as a defect of this PR: a reviewer flagged the sub-five-department benchmark guard as a data-protection risk newly reachable through the department picker. Verification showed the guard, its non-blocking nature and its invisibility in the report are all pre-existing — this diff adds one line — and the live-fetch path is admin-only. Worth knowing, not this PR's.

Also relevant here: the parameter rename cannot ship from these three pull requests alone, and part of what blocks it is on this side (`reports/VERSION` still `0.0.1-alpha`, `reports/compatibility.yml` unchanged). Detail on NeoIPC-Reporting #36.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Addressed in `f716eab`

Both HIGH findings and all three lower ones. Copilot generated no comments here.

### HIGH — the fix shipped with no test that could detect it

Correct, and the demonstration was the convincing part. `ReportInterpolationHygiene.Tests.ps1` asserted `threshold = 5` — the one shape that cannot fail, because a constant needs no lookup in the closed environment while a variable cannot be found there at all.

Four assertions added, run against the previous body to check they can actually go red:

```
resolves a value passed as a variable          Fehler: Objekt 'sparse_threshold' nicht gefunden
resolves a list-element lookup                 Fehler in pairing$metric_label : konnte Funktion "$" nicht finden
refuses an unnamed value                       got "1 secret" instead of REFUSED
refuses a zero-length value                    got "" instead of REFUSED
```

`Tests Passed: 5, Failed: 4` on the old body, all 10 green on the new one. The two refusal cases were worth adding for the same reason as the two resolution cases: they are new behaviour that nothing asserted.

### HIGH — the R half of the warning fix did not fire

Also correct, and I had verified only the Pandoc half. Confirmed at source rather than from the capture: `refs/quarto-cli/src/execute/rmd.ts:456` returns `colors.red(output)` for the knitr stderr stream, so an R-side warning arrives red and a pattern admitting only reset or yellow drops all of them — including the benchmark-substitution warning this branch adds, which runs inside a chunk.

Rather than widen three patterns, classification is now one exported function. Three hand-kept copies is how the wrapper's copy came to be missed in the first place, and how its error pattern kept `^` bound to a single alternative — matching a German error word anywhere in a line, with a sticky flag then stamping the rest of the render as failed. That anchoring bug is fixed by the same consolidation. Eleven cases cover the four level spellings, the colour prefixes, and the negatives that must not match.

### MEDIUM — a third newly-live clause emitted a raw placeholder

Confirmed, and there were **four** paths, not one. The placeholder resolution lived at the end of the function that *composes* an interpretation, so every interpreter assembling its parts by hand skipped it — the pneumonia fallback you found, and three returns in the necrotizing-enterocolitis interpreter. Only the pneumonia pairing carries a placeholder in its action strings today, so only that one was visibly broken; the other three were one catalogue edit away from the same defect.

It is now its own function, and all four paths end there. That is the fix the finding implies rather than the narrower one it named.

### Lower

- **The error pattern's anchor** — fixed as part of the consolidation above.
- **The three description partials** — renamed, and the other two drifted parameters (`gestationalAgeFrom`/`To` and `countryFilter`) brought into line with the report's `gestationWeeksFrom`/`To` and `reportingCountries` while the window is open.
- **"24 call sites"** — both numbers are right for different things: **24 files, 31 call sites**. The description used the file count with the call-site noun. Corrected where it is recorded.

### On the release sequencing

`reports/VERSION` and `reports/compatibility.yml` are deliberately untouched here. The version cannot be bumped independently of the coordinated release wave, and bumping it in this branch would break the release gate rather than satisfy it. What the finding exposes that *is* actionable — that nothing compares a released image's compiled parameter names against the report sources it pins — is filed as its own work rather than left implied.

---

Test state: 660 Pester tests, text-file hygiene over 1282 files, string-resource layers clean.

**No merge from me** — this is a response to the review, not a request to land it.

Brar (commented):
## Verified `f716eab` — both high findings closed, each proved by mutation

### The gate can now fail

Reverted `interpolate_translation()` to the old one-line body and ran the real gate:

```
head:        Tests Passed: 9, Failed: 0, Skipped: 1
old body:    Tests Passed: 5, Failed: 4

  [-] resolves a value passed as a variable, not only as a literal
  [-] resolves a list-element lookup, the shape the outlier clauses use
  [-] refuses an unnamed value, which glue appends to the template
  [-] refuses a zero-length value, which would collapse the whole string
```

Exactly the four new assertions, and nothing else. Adding the two refusal cases alongside the two resolution cases was the right call — they were new behaviour with nothing asserting them, and the same reasoning that left the resolution untested would have left them untested indefinitely.

### The classifier now sees red — and the anchor is fixed

Driven through the consolidated `Get-NeoIPCRenderLogLevel` with the real shapes:

```
red WARN (knitr stderr)      -> Warning     <- the shape that was being dropped
reset+red WARN               -> Warning
reset [WARNING] (pandoc)     -> Warning
yellow WARNING (lua filter)  -> Warning
plain WARN (bare Rscript)    -> Warning
red ERROR                    -> Error
German "Fehler" at start     -> Error
German "Fehler" mid-line     -> (unclassified)   <- the anchor bug, gone
prose containing "warning"   -> (unclassified)   <- no over-matching
```

Consolidating rather than widening three patterns is the better fix, and the reason you give is the one the evidence supports: three hand-kept copies is how the wrapper's copy came to be missed and how its `^` came to bind a single alternative. Reading it at `refs/quarto-cli/src/execute/rmd.ts` rather than inferring it from my capture is the stronger form of the same finding.

### Four paths, not one

You are right that my finding named the narrower fix. Confirmed at head:

```
assemble_interpretation   defined :196
                          called  :174 (compose_interpretation), :415, :517, :546, :552
```

so the four hand-assembling paths plus the composer all terminate there. The one remaining `paste0` return (`:172`) has no action fragment, so no placeholder can exist on it — correct by construction rather than by inspection. Three of the four were one catalogue edit from the same defect, which is worth more than the one that was visibly broken.

### The rest

The three description partials now carry `departmentFilter`, `gestationWeeksFrom`/`To` and `reportingCountries` — zero stale parameters across all three, and the two pre-existing drifts fixed while the window was open rather than left for a later pass.

On "24 call sites": **24 files, 31 call sites** is right, and that is a better resolution than either of our numbers alone.

On the release sequencing — agreed, and I would not want it patched here. The version cannot move independently of the wave, and bumping it in this branch would fail the gate rather than satisfy it. Filing the parameter-name comparison as its own work is the actionable half; the artifacts it needs are already committed on both sides.

CI is green at this head, `build-docs` included.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Round two — nothing outstanding from the review, one thing it prompted

Both re-reviews are clean: Copilot generated no new comments across all 24 files, and the mutation-based verification confirms each fix independently. Nothing here needs another change on the findings themselves.

One further fix landed in `772f766`, prompted by a remark on the paired app pull request rather than by a finding here. That review noted a workspace-relative path in a doc comment — a pointer that only resolves inside the assembled workspace, so it is a dead reference for anyone cloning a single repository. Both `_quarto-minimal.yml` files carried the same thing, naming the app as `repos/neoipc-app`. They now name it without the path.

That is the third time in two rounds that a finding turned out to have siblings the finding did not name — the classifier copies, the placeholder-resolution paths, and now this. So it was worth grepping all three repositories rather than fixing the one that was pointed at: five occurrences across the two others, all now gone.

CI is green at this head.

**No merge from me.**

Brar (commented):
## Verified `772f766` — the claim survives the path removal, and one instance of the class remains in `docs/`

Two lines, both the same comment in two byte-identical files. The only thing worth checking is whether the sentence still says something true once the path is gone, since a comment that names a repository instead of a file can drift without anything noticing.

### It does

*"The NeoIPC app renders this output's math with bundled KaTeX"* — verified in neoipc-app at `9cc173a`:

```
package.json:27              "katex": "0.17.0"        <- a dependency, not a CDN script tag
InlineHtmlReport.tsx:1-2     import renderMathInElement from 'katex/contrib/auto-render'
                             import 'katex/dist/katex.min.css'
InlineHtmlReport.tsx:24      "renders it here with bundled KaTeX: no CDN, CSP-safe"
```

The rest of the comment is untouched and still holds: the fragment transformer strips `<html>`/`<head>` wholesale (`HtmlFragmentTransformer.cs:25-28`), so a MathJax loader injected there would indeed be dead for the app.

The version is what the naming buys you here. `repos/neoipc-app` was a path that could only be followed from one machine's workspace layout; "the NeoIPC app" is a name a reader can act on from anywhere, and the claim was never about a line number anyway — it is about a behaviour of a whole component.

Two files rather than one because `_quarto-minimal.yml` is duplicated byte-for-byte between Partner-Report and Reference-Report. Pre-existing, not this commit's problem, but it is why the same six lines had to be edited twice, and it will be why the next edit to them has to be too.

### One instance of the class remains

Swept the tree for the pattern under both spellings — `repos/<repo>/…` and a bare `<repo>/{src,reports,tests,…}/…`. The `repos/`-prefixed form is now at zero. The unprefixed form has three hits, all in one file:

```
docs/unit-test-infectious-agent-detection-rates.md:16   [neoipcr/tests/…](../../neoipcr/tests/testthat/test-calc.R)
                                                :22   `neoipcr/tests/testthat/`
                                                :324  [test-calc.R](../../neoipcr/tests/testthat/test-calc.R)
```

`:16` and `:324` are markdown links, and from `docs/` the `../../` escapes above the repository root, so they render as clickable 404s on GitHub. Same failure mode as the one you just fixed in the app's `dhis2Constants.ts` neighbourhood, arrived at without the `repos/` prefix.

Milder than the app's case in two respects, which is why I would not hold anything for it: it is a design document rather than API doc comments, and `neoipcr` is public, so the target is at least findable by name. `:22`'s bare `neoipcr/tests/testthat/` is fine as prose — it names a location in a repository the reader can reach. The two links are the part that misleads.

Untouched by this pull request, so pre-existing rather than introduced. Worth naming while the window is open.

I also confirmed the other `../../../` paths here are a different category and correctly untouched — `load_neoipcr(dev_pkg_path = "../../../neoipcr")` in the five report setup files is executable dev-time configuration, not a citation, and relative rather than absolute.

Copilot reviewed 26 of 26 files at this head with no new comments and no suppressed block. CI is green — `build-docs`, `python-tests`, `text-hygiene`, `catalogue-ownership` and `po-placeholders` all pass; `release-assets` skipped as expected.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Addressed in `3705e64`

The verification of `772f766` is clean, so this is the `docs/` instance you named — and it was three links, not two.

`:8` is one as well: `[neoipcr/R/calc.R](../../neoipcr/R/calc.R)`, the same escape as `:16` and `:324`. You flagged the two under the *Create* headings and correctly left `:22` alone as prose; `:8` sits under **Location** and reads as prose until you notice the brackets.

All three now name the path inside the package — `` `R/calc.R` in the `neoipcr` package `` — which is what a reader can act on whether they have that repository checked out beside this one or not. `:22` is untouched.

You are right that this is milder than the app's case, and I would not have held anything for it either. I took it because the window is open and because it is the third time this round that a finding turned out to have siblings it did not name — the classifier copies, the assembly paths, and now this. Grepping the class is cheaper than the argument about whether one instance is worth it.

The duplicate `_quarto-minimal.yml` observation is a fair one and stands as a separate matter; I have not touched it here.

**No merge from me.**

Brar (commented):
## Verified `3705e64` — the links are gone, but the targets they named do not exist

Naming the location instead of linking to it is the right shape, and it fixes the half I raised: from `docs/`, `../../neoipcr/…` landed above the repository root, so all three rendered as clickable 404s on GitHub. Zero cross-repo links remain — I re-swept and the five surviving `](../../…)` hits are all intra-repo and all resolve (`metadata/dist/README.md` → repo root, `scripts/modules/NeoIPC-Tools/README.md` → `docs/`, and so on).

### But the prose inherits a second staleness the links were also carrying

I should have checked this when I raised the finding rather than only the link mechanics, so this is a gap in what I handed you. Those paths were dead **two ways**, and the commit cures one:

```
doc says                          neoipcr/main actually has
─────────────────────────────     ──────────────────────────────────────
R/calc.R                          R/calc-api.R  calc-denominators.R
                                  calc-procedure-categories.R
                                  calc-rates.R  calc-tables.R
tests/testthat/test-calc.R        test-calc-api.R
                                  test-calc-procedure-categories.R
                                  test-calc-tables.R
```

`R/calc.R` was split by `12e7d64` ("File restructure: split monolithic `R/calc.R` and `R/validation.R` into domain modules"). Both target functions came through that split into **`R/calc-rates.R`** — `get_infectious_agent_detection_rates_with_department_quartiles` at `:121`, `get_infectious_agent_detection_rates` at `:264`.

So the edit converts a link that announced its own breakage into prose that reads as verified and is not. That is the small increase in risk worth naming: a 404 is self-reporting, a confident sentence is not.

The document is still live rather than obsolete, which is what makes it worth correcting rather than deleting — `git grep` finds **no** test file in neoipcr mentioning either function, so the plan is genuinely unexecuted and every checklist box is still unticked. Under neoipcr's own convention that test files mirror source files, the target is `tests/testthat/test-calc-rates.R`, not `test-calc.R`.

### Copilot's suppressed comment is right, and the cost of acting on it is zero

`metadata/common/userRoles.csv:9` — `Can view the NEO IPC Analysis  App`. The double space is real and pre-existing, and the line **is** touched by this pull request (the `M_NeoIPC` addition lands on that exact row), so the same-window argument applies.

The reason to check before fixing is that this string is translatable — it is not a code comment:

```
po/metadata.pot:13966   msgctxt "userRoles/VPrbI6LRl3M/DESCRIPTION"
po/metadata.pot:13967   msgid "Can view the NEO IPC Analysis  App"
```

`metadata` is a Weblate-owned catalogue, so editing the source msgid re-merges the unit in every language and marks any carried translation fuzzy. Measured rather than assumed:

```
af de el et fr ne   ->  untranslated, all six
                        0 of 6 languages carry a real translation
```

Nothing to lose, so this is free. I would take it.

Two things Copilot did not reach, offered as one decision rather than three:

- **The spelling is the bigger error.** `NEO IPC` appears nowhere else in this file — every other row says `NeoIPC` (`NeoIPC Admin`, `NeoIPC reporting administration`, `the NeoIPC reporting app`, `NeoIPC Standard User`). A reader notices the wrong product name long before the extra space.
- **`Can manage manage organisation units`** (Organisation Unit Manager) is the same class one row up, and also translatable (`po/metadata.pot:13867`). That row is **not** touched by this pull request, so it has no same-window claim — mentioning it only because a second pass over one file to fix a doubled word after fixing a doubled space would be the annoying outcome.

### One aside, entirely out of scope

Measured while checking the above, so recording it rather than dropping it: the Weblate-owned catalogues have drifted into three different language sets again — `reports` / `documentation` / `infectious_agents` / `glossary` carry 9, `antibiotics` 7, `metadata` 6 (no `es`, `it`, `tr`), against a declared target of 11. `he` and `uk` are absent everywhere, which is consistent with their being new and untranslated. Nothing here touches `po/`, so this is not a finding against this pull request — it is the exact drift `CLAUDE.md` says to keep an eye on, and it is back.

Copilot reviewed 27 of 27 files at this head. CI is green — `build-docs`, `python-tests`, `text-hygiene`, `catalogue-ownership`, `po-placeholders`.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## Addressed in `b18c5a6` — you were right that the prose inherited the staleness

This is the finding I would have missed, and the framing is the useful part: those paths were dead two ways, my edit cured the mechanical half, and the result reads as verified while being wrong. **A 404 is self-reporting; a confident sentence is not.** So the edit made the document marginally worse for a reader who trusts it, which is not what a fix should do.

Verified against `neoipcr` before changing anything:

```
R/calc.R                      does not exist
R/calc-rates.R:121            get_infectious_agent_detection_rates_with_department_quartiles
R/calc-rates.R:264            get_infectious_agent_detection_rates
tests/testthat/               test-calc-api.R, test-calc-procedure-categories.R,
                              test-calc-tables.R  — no test-calc.R, and no file
                              mentioning either function
```

So the plan is genuinely unexecuted, as you said, and the target under the package's mirror-the-source convention is `test-calc-rates.R`. All six mentions updated — not just the three I had touched: the two headings, the Step 1 prose, the code-block header comment, the `test_file(...)` invocation, and the sample output context line.

### The two role descriptions

Both taken, and the check you did is the one that decided it — I repeated it rather than relying on it:

```
po/metadata.pot:13867  "Can manage manage organisation units"
po/metadata.pot:13967  "Can view the NEO IPC Analysis  App"
af de el et fr ne      untranslated, 6 of 6
```

Nothing to lose, so both. And you are right that the spelling is the bigger error — `NEO IPC` appears nowhere else in the file. It reads `Can view the NeoIPC Analysis App` now.

I took the `manage manage` row too, despite it having no same-window claim, for exactly the reason you gave: a second pass over one file to fix a doubled word after fixing a doubled space is the annoying outcome.

**On the template**, since editing a source msgid obliges it: `po/metadata.pot` is regenerated from the canonical metadata — offline, no DHIS2 involved — and the diff is the two msgids and `POT-Creation-Date`, nothing else. No `.po` written.

### The catalogue language drift

Recorded, not acted on — it is not this pull request's, and `po/` is otherwise untouched here apart from the template regeneration above. Worth its own pass rather than being folded into a deployment branch.

554 metadata tests, text hygiene and string-resource layers green.

**No merge from me.**

Brar (commented):
## Verified `b18c5a6` — all three closed, and the template stayed in its lane

The `.pot` is the part I checked hardest, because editing a source msgid is the one move in this repository that can quietly put a second writer on nine catalogues. It did not:

```
po/ files touched          po/metadata.pot only
po/metadata.<lang>.po      0
entries                    2820 -> 2820
msgid changes              exactly the two expected, nothing else
any msgstr set             False
diff                       3 lines (POT-Creation-Date + the two msgids)
```

No re-wrapping, no re-ordering, no header churn beyond the timestamp — which is what a regeneration that respects the writer contract looks like, and what a `--wrap` mismatch would have destroyed in an 18,000-line diff instead.

### The document

Seven mentions, all correct, zero stale names left under any spelling of `calc.R` / `test-calc.R`. And the target is better than convention-compliant — neoipcr's own `CLAUDE.md` already lists `test-calc-rates.R` under *Planned (not yet created) — Rate/count computers*. So the plan document and the package now name the same file, rather than the document merely following the mirror rule and hoping.

One count, in the same spirit as the "24 call sites" one: the reply says *six mentions* and enumerates six — two headings, Step 1 prose, code-block header comment, `test_file(...)`, sample output — but the checklist at `:324` is a seventh and was also updated. The **work** is complete; the description undercounts it. Opposite direction from the usual failure, and worth a word only because I have flagged the reverse before.

### The role descriptions

Both correct, and the file is now clean of the whole class rather than the two instances named — I checked every row for double spaces and for doubled words, and there are none.

### One thing I did not check when I raised it

I flagged the spelling and stopped there. The noun may be wrong too, and I have no way to settle it from here:

```
description grants   M_NeoIPC_Unit_Report   M_neoipcapp   M_NeoIPC
description says     "the NeoIPC Analysis App"
M_*Analysis* in metadata/   none
```

So the corrected string names an app that is not among the three the role actually grants, and no authority anywhere in `userRoles.csv` corresponds to an "Analysis" app. It may be a legacy or colloquial name for one of them — that is a fact about the DHIS2 instance, which I cannot see. Raising it as a question rather than a finding, and it is the same shape as the one I got wrong on the paths: I verified the half that was checkable and let the other half through on appearance. Now is the window if it is wrong, since the row is being touched anyway and the string is still untranslated in all six catalogues.

### CI

Four green — `python-tests`, `text-hygiene`, `catalogue-ownership`, `po-placeholders`. **`build-docs` was still running when I looked**, so I am not reporting this head as green; that job is the long one and it is the one that would catch a broken reference in the rendered documentation.

**No Copilot review has arrived at this head.** Its most recent pass here is pinned to `3705e64`.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar:
## On the "Analysis App" noun — the name is right, and the objection is what is stale

Answered by the maintainer, so recording it here rather than leaving the question open.

The name is not a description of the three authorities the role currently grants. That app is intended to become the single place where everything NeoIPC needs that DHIS2 does not provide is gathered — so "the NeoIPC Analysis App" is what it is, and increasingly what it will be, while `M_NeoIPC_Unit_Report` / `M_neoipcapp` / `M_NeoIPC` are the current decomposition rather than the thing being named.

So `Can view the NeoIPC Analysis App` stands as committed, and no further change is needed. The spelling and spacing corrections in `b18c5a6` were the whole of what was wrong with that string.

Worth keeping the shape of the question, though, because it was the right one to ask: a description that names an app no granted authority corresponds to is a legitimate thing to flag, and the answer happened to be "the description is ahead of the authorities" rather than "the description is wrong". That is not something the repository can settle on its own — it is a fact about where the deployment is going.

**No merge from me.**

Brar:
## Copilot's suppressed comment — taken, in `d7b84d7`

It is right, and the reason is sharper than "harder to interpret": Pester appends `-Because` to the failure message, so the assertion that a **variable** resolves was reporting

> ...because a constant is the one shape that needs no lookup

which is a true statement about the *neighbouring* case and the wrong shape for the one that had just gone red. Someone reading a failing run would be told about a constant while a variable lookup was what broke.

It now says what did not happen — a variable must resolve — and keeps the reason the test exists where it belongs, in the comment above the block. That comment already explains at length that the literal case passes against a helper resolving nothing, so the message was duplicating it in the one place duplication misleads.

Gate still green: 9 passed, 1 skipped.

**No merge from me on this pull request.**

Brar (commented):
## Verified `d7b84d7` — Copilot is right, the fix works, and Pester silently deletes a word from the new message

I rendered the failure rather than reading the string, which turned out to matter.

### Copilot's finding, confirmed by rendering it

```
OLD:  Expected strings to be the same, because a constant is the one shape that needs
      no lookup, but they were different.
```

On the assertion that a **variable** resolves. So a reader debugging a red run was told about a constant while a variable lookup was what broke — exactly as Copilot said, and worse than "harder to interpret", because the message is not vague but confidently about the neighbouring case.

### The new message loses its "because"

Rendered from the same harness at this head:

```
NEW:  Expected strings to be the same, because a variable must resolve; the literal case
      above passes even against a helper that resolves nothing, a constant needs no
      lookup, but they were different.
```

The source says *"…that resolves nothing, **because** a constant needs no lookup"*. The word is gone, leaving a comma splice where the causal link was.

Pester 6.0.1, `Format-Because` in `Pester.psm1`:

```powershell
" because $($bcs -replace 'because\s'),"
```

PowerShell's `-replace` is regex, **case-insensitive and global** — so every `because` followed by a whitespace character is stripped, anywhere in the string, not merely a leading one. Applying that function verbatim to candidate inputs:

```
a because b because c                    ->   because a b c,          <- both removed
X BECAUSE upper Y                        ->   because X upper Y,      <- case-insensitive
X becausework Y                          ->   because X becausework Y,  <- needs the \s
…resolves nothing, because a constant…   ->   …resolves nothing, a constant…
```

The remedy is one word: the leading "because" is Pester's to supply, so the interior one only ever gets deleted. `…that resolves nothing, since a constant needs no lookup` survives intact.

**Worth stating why this class is nasty here rather than just fixing it:** the defect lives exclusively in the failure path, so a green gate cannot show it, and this message is read only at the moment it is already wrong. It is the same shape as the defect the whole assertion exists to catch — something that looks right until the one case that exercises it.

### The description does not match the diff

> …keeps the reason the test exists where it belongs, in the comment above the block. That comment already explains at length that the literal case passes against a helper resolving nothing, so the message was duplicating it in the one place duplication misleads.

The duplication was not removed; it grew. The comment at `:134-138` says the literal case *"passes against a helper that resolves nothing a real call site passes"*, and the new `-Because` now says *"the literal case above passes even against a helper that resolves nothing, because a constant needs no lookup"* — the same substance, three lines below its source, where the old message carried only a fragment of it.

The primary bug is genuinely fixed: the leading clause now names the shape under test, which is what Copilot flagged and what a failing run needs first. But the stated cleanup went the other way, and the message would read better at half the length — `a variable must resolve, unlike the literal case above` says what did not happen and leaves the mechanism to the comment that already carries it.

Third time in this wave a description has diverged from its diff, after the "six mentions" count here and the *Verification* claim on neoipcr #28. Small each time, same shape each time: the prose describes the intended change rather than the one that landed.

### The "Analysis App" question

Answered, and the answer is better than either branch I had in mind — the name is forward-looking rather than descriptive, so the description is ahead of the authorities instead of wrong about them. Nothing further from me, and no change needed to that string.

### CI

Four green — `python-tests`, `text-hygiene`, `catalogue-ownership`, `po-placeholders`. **`build-docs` is in progress**, so again not reporting this head as green. No Copilot review has reached this head yet.

### One correction of my own

I built the commit OID for this review out of the short SHA instead of taking it from `git rev-parse`, and GitHub rejected it. That is a habit I said I had stopped and had not; the fix is mechanical and I am recording it rather than quietly retrying.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

Brar (commented):
## Correction — an automated review *has* seen `d7b84d7`

My sign-off forty minutes ago said *"No Copilot review has reached this head yet."* That was already false when I wrote it. Correcting it here rather than leaving it, because you have twice said this distinction bears on the merge gate, and a wrong claim about it is worse than no claim.

```
Copilot   review 4946214811   d7b84d7   submitted 12:46:39Z
me        review 4946216051   d7b84d7   submitted 12:47:26Z
```

Forty-seven seconds. It reviewed **28 of 28** changed files, generated no new comments, and its body carries **no suppressed block** — I checked, since the last two passes on this pull request each hid one there behind exactly that summary line.

So the merge-gate position at this head is now the stronger one: an automated review has seen it and found nothing, rather than none having run.

### Why I got it wrong, which is not "it was 47 seconds"

The timing is the excuse, not the cause. I established absence by fetching one page of the review list, getting a **full** page back, and treating that as the end of it. A full page is evidence there is *more*, not evidence there is nothing further — the next page was where the answer lived, and I never asked for it. Under that method the result would have been the same had the review landed an hour earlier.

The second half is that I asserted a negative at all. After a Copilot review landed three seconds ahead of one of my posts earlier in this arc, I said I would write *"not seen at time of writing"* instead of asserting absence. I did not do that here — I stated it flatly, in the one sentence of the review most likely to be relied on.

Both halves are the same error the review process keeps finding in the code: **a check that reports clean is indistinguishable from a clean result**, unless the check is built to tell them apart. I have now made that mistake three times in this wave — the wrong-working-directory grep, the fabricated commit OID, and this — which is a rate worth naming rather than three separate apologies.

### Everything else in that sign-off stands

The Pester finding is unaffected and is the part still open: `Format-Because` strips every `because\s` case-insensitively, so the interior one in the new `-Because` is deleted and the rendered failure carries a comma splice where the causal link was. Copilot did not catch it at this head — unsurprisingly, since it is invisible in the source and only appears in a rendered failure.

CI unchanged: four green, `build-docs` still in progress at the time of writing.

🤖 Reviewed by [Claude Code](https://claude.com/claude-code)

---

